Issue #367, opened 2026-05-30
test: MapImageGenerator coordinate math and bounds logic are uncovered


dmccoystephenson:
## Summary

`src/mapimage/mapImageGenerator.py` contains the only non-trivial geometry in the minimap subsystem and has **no test of its own**. A grep for `MapImageGenerator` across `tests/` returns only mocked references inside `tests/mapimage/test_mapImageUpdater.py`; the class is never instantiated or exercised directly.

Untested logic includes:

- `pasteRoomImagesAtCorrectCoordinates` (`src/mapimage/mapImageGenerator.py:62`) — parses room coordinates from a filename (`roomImageFilename.split(".")[0].split("_")`) and computes pixel positions:
  ```python
  picX = int(self.mapImageSizeInPixels / 2) + roomX * roomSize - int(roomSize / 2)
  picY = int(self.mapImageSizeInPixels / 2) + roomY * roomSize - int(roomSize / 2)
  ```
- The four-sided in-bounds check that decides paste-vs-skip (`picX >= 0 and picY >= 0 and picX < self.mapImageSizeInPixels and picY < self.mapImageSizeInPixels`).
- The `percentUpdated` calculation over `(self.numRoomsInEachDirection * 2 + 1) ** 2`.

## Why it matters

These are exactly the kind of arithmetic + boundary conditions where off-by-one and sign errors hide (negative `roomX`/`roomY`, rooms near the map edge, the origin room). A regression here silently mis-places or drops rooms on the minimap with no error, and nothing in the suite would catch it.

## Suggested fix

Add `tests/mapimage/test_mapImageGenerator.py` following the existing per-subdir naming convention. Cover at minimum: coordinate parsing from a `x_y.*` filename; `picX`/`picY` for the origin and a few offset rooms; the in-bounds boolean at and just past each boundary; and an out-of-bounds room being skipped (asserting it is not pasted). The PIL `Image` and directory listing can be mocked the way `test_mapImageUpdater.py` already mocks collaborators.

_Filed by Claude during an automated triage pass; the absence of direct coverage was confirmed by grepping the test tree._

Issue #365, opened 2026-05-30
config.yml is out of sync with the code: 2 keys are read but never used, 2 keys the code reads are missing


dmccoystephenson:
## Summary

During an automated triage pass, `config.yml` was compared against the keys the configuration loader (`src/config/config.py`) actually reads and the rest of the code actually consumes. Four mismatches were confirmed by reading both sides.

**Keys documented in `config.yml` but never used by any code:**

- `config.yml:7` — `black: [0, 0, 0]`
- `config.yml:8` — `white: [255, 255, 255]`

  These are parsed into attributes at `src/config/config.py:178-179`:
  ```python
  self.black = self.getColorValue(configValues, "black", (0, 0, 0))
  self.white = self.getColorValue(configValues, "white", (255, 255, 255))
  ```
  A grep across `src/` (excluding the definitions and vendored `src/lib/`) finds **zero** references to `self.config.black` / `self.config.white`. The places that need those colors use hardcoded literals instead (e.g. `src/screen/worldScreen.py:1428` `white = (255, 255, 255)`, `src/ui/energyBar.py:40-43`). The two config keys are dead.

**Keys the code reads but that are absent from `config.yml`:**

- `cropGrowthTicks` — read at `src/config/config.py:201-202` (default `1800`) and consumed at `src/world/room.py:301` (`if tick - entity.getTickPlanted() >= config.cropGrowthTicks:`). Not present in `config.yml`, so the crop-growth timing is undiscoverable/untunable from the config file.
- `pushableStone` — read at `src/config/config.py:230` (default `True`) and consumed at `src/screen/worldScreen.py:433`. It is even exposed as an in-game toggle at `src/screen/configScreen.py:66` (`("Pushable Stone", "pushableStone")`), yet it is the only such in-game toggle missing from the `# dynamic (can be changed in game)` section of `config.yml` (which lists `fullscreen`, `autoEatFoodInInventory`, `removeDeadEntities`, `showMiniMap`, `cameraFollowPlayer`, `limitTps`, `dayNightCycleEnabled`).

## Why it matters

The config file is the user-facing surface for tuning the game. Documenting two keys that have no effect is misleading, and omitting two keys the code honors hides real tunables — `pushableStone` especially, since it is a toggle the player can flip in-game but cannot pre-set or discover in the file.

## Suggested fix

[...]

Issue #364, opened 2026-05-30
test: add unit tests for WorldScreenPersistence (save/load currently uncovered)


dmccoystephenson:
## Summary

`src/screen/worldScreenPersistence.py` handles persistence of player location, player attributes, player inventory, and rooms (JSON I/O plus schema validation), but it has no test coverage — a grep of `tests/` for `WorldScreenPersistence` returns zero references.

This is core game-state persistence; an undetected regression here means lost or corrupt saves.

## Suggested coverage

Add `tests/screen/test_worldScreenPersistence.py` exercising:
- Round-trip `save -> load` for player location and player attributes (values preserved).
- Room path construction (`buildRoomPath`) for representative coordinates.
- Behavior when the target file is missing or corrupt.

Use the shared DI container from `tests/conftest.py` (prefer `resolve(Type)` / `override_dependency`) and point `pathToSaveDirectory` at a `tmp_path` fixture so tests don't touch real saves.

_Filed by Claude during an automated triage pass; the absence of coverage was confirmed against `tests/`._

dmccoystephenson:
Note from a later triage pass: this issue's suggested coverage references `buildRoomPath`, which was removed as dead code (zero callers) in PR #380 when the room save-file path was consolidated into `Config.getRoomFilePath`. The remaining coverage value still stands — `WorldScreenPersistence` (player location / attributes / inventory round-trips, and missing/corrupt-file behavior) is genuinely untested — but a future PR taking this on should target the methods that actually exist rather than `buildRoomPath`. Reported by Claude.

Issue #363, opened 2026-05-30
Save: schema validation failures are logged but invalid data is still written to disk


dmccoystephenson:
## Summary

In two JSON writers, a `jsonschema` validation failure during save is caught and logged, after which the **invalid** data is written to disk anyway — so validation provides no actual protection against persisting corrupt saves.

- `src/inventory/inventoryJsonReaderWriter.py` (`saveInventory`): `jsonschema.validate(...)` is wrapped in a `try/except` that logs the `ValidationError`, then `json.dump(toReturn, f)` writes the data regardless.
- `src/codex/codexJsonReaderWriter.py` (`save`): identical pattern — validation error is logged, then the data is written.

## Inconsistency

Other writers handle the same situation differently — `src/stats/stats.py`, `src/screen/worldScreenPersistence.py` (player location / attributes), and `src/world/tickCounter.py` do **not** wrap `validate()`, so a validation failure there raises (and can crash on save/load) rather than silently persisting. The codebase currently handles validation failure three different ways.

## Why it matters

This is a single-player game with on-disk player saves. Silently writing schema-invalid inventory/codex JSON can corrupt a save that may then fail or partially load on next launch.

## Suggested fix

Adopt one consistent policy across all JSON writers. For save integrity, on validation failure **do not overwrite the existing good file** — either abort the write (and log/surface it), or write to a temp file and atomically `os.replace()` only after validation passes.

_Filed by Claude during an automated triage pass; the two code paths above were read and confirmed._

PR #379, opened 2026-06-06
refactor: extract shared config-file rewrite loop into Config helper

### LINKED ISSUE #366: refactor: config-file rewrite loop is duplicated verbatim between config.py and keyBindings.py
## Summary

Two methods that persist values back into the on-disk config file contain a read → update-matching-keys → append-new-keys → write-back loop that is **character-for-character identical** apart from how they build the `savedValues` dict.

- `src/config/config.py:243` — `saveWindowSize(self, width, height)`
- `src/config/keyBindings.py:129` — `saveToConfigFile(self, config)`

Both run the same body:
```python
configFilePath = ...getConfigFilePath()
lines = []
if configFilePath.exists():
    try:
        lines = configFilePath.read_text(encoding="utf-8").splitlines()
    except (OSError, UnicodeDecodeError):
        lines = []

# savedValues built differently in each method

updatedKeys = set()
newLines = []
for line in lines:
    stripped = line.strip()
    if stripped == "" or stripped.startswith("#"):
        newLines.append(line)
        continue
    parts = stripped.split(":", 1)
    if len(parts) == 2:
        key = parts[0].strip()
        if key in savedValues:
            newLines.append(key + ": " + savedValues[key])
            updatedKeys.add(key)
            continue
    newLines.append(line)

for key, value in savedValues.items():
    if key not in updatedKeys:
        newLines.append(key + ": " + value)

try:
    configFilePath.write_text("\n".join(newLines) + "\n", encoding="utf-8")
except OSError as e:
    _logger.warning(...)
```

The only real difference is `savedValues` construction (window dimensions vs. prefixed keybindings) and the log message.

## Why it matters

Any fix or change to the config-file rewrite behavior (comment handling, quoting, encoding, atomic write) must currently be made in two places and is easy to drift. It is also the kind of duplication that invites a bug fixed in one copy but not the other.

## Suggested fix

Extract the shared loop into a single helper — e.g. `Config._writeKeyValues(self, savedValues: dict[str, str])` on the `Config` class (it already owns `getConfigFilePath()`), and have both `saveWindowSize` and `KeyBindings.saveToConfigFile` build their `savedValues` dict and delegate to it.

_Filed by Claude during an automated triage pass; both methods were read in full and confirmed identical apart from the `savedValues` construction._

dmccoystephenson:
## Summary

DRY refactor consolidating the duplicated config-file rewrite logic into a single source of truth.

- `Config.saveWindowSize` and `KeyBindings.saveToConfigFile` contained a **character-for-character identical** read → update-matching-keys → append-new-keys → write-back loop, differing only in how `savedValues` is built and the log message.
- Extracted the loop into `Config._writeKeyValues(self, savedValues, errorMessage)` (`src/config/config.py`). Both callers now build their `savedValues` dict and delegate; the differing warning text is threaded through `errorMessage` so existing log behavior is preserved.
- Removed `keyBindings.py`'s now-orphaned `getLogger` import and module-level `_logger` — its only use was the warning that moved into `Config`.

Net **−13 LOC** across 3 files; `keyBindings.saveToConfigFile` shrank from ~42 lines to ~8.

## Why

Any future change to the rewrite behavior (comment handling, quoting, encoding, atomic write) previously had to be made in two places and was easy to drift — the classic "bug fixed in one copy but not the other" trap.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 461 passed (was 460; +1 new)
- [x] **Regression coverage:** the existing `test_config.py` (`saveWindowSize`) and `test_keyBindings.py` (`saveToConfigFile`) suites exercise the extracted helper through both public callers — update-in-place, append, whitespace-before-colon, comment preservation, round-trip — and all still pass unchanged.
- [x] Added `test_write_key_values_updates_appends_and_preserves` exercising the helper directly (update + append + comment/blank-line preservation in one call).
- [x] Formatting run scoped to the 3 changed files only (not tree-wide `format.sh`), so no unrelated drift entered the diff.

- **#373** (room save-file path duplicated 5×) — the other refactor candidate; deferred because it spans 4 files including a persistence path (higher blast radius) and pairs naturally with the #370 atomic-write work, so it warrants its own focused PR.
- **#362** (copilot-instructions.md staleness), **#365** (config.yml drift) — touch do-not-auto-merge paths; better in their own PRs.
- **#370, #363** (save-corruption / schema-write bugs) — larger; need atomic-write design and dedicated regression tests.
- **#372, #367, #364, #368** (coverage / robustness) — out of this PR's theme.

Closes #366

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,13 @@ logged in detail below.
 
 ## AI Agent Sessions
 
+### 2026-06-06 — Deduplicate the config-file rewrite loop (issue #366)
+- **`src/config/config.py`:** Extracted the read → update-matching-keys → append-new-keys → write-back loop into a new `Config._writeKeyValues(self, savedValues, errorMessage)` helper. `saveWindowSize` now just builds its `savedValues` dict and delegates; the differing log message is passed through `errorMessage` so existing log behavior is preserved.
+- **`src/config/keyBindings.py`:** `KeyBindings.saveToConfigFile` now builds its prefixed `savedValues` dict and delegates to `config._writeKeyValues(...)`, dropping the verbatim-duplicated loop (~37 lines). Removed the now-unused `getLogger` import and module-level `_logger` (the only use was the warning that moved into `Config`).
+- **`tests/config/test_config.py`:** Added `test_write_key_values_updates_appends_and_preserves`, a direct test of the helper covering in-place key update (no duplicate), new-key append, and comment/blank-line preservation in a single call. The existing `saveWindowSize`/`saveToConfigFile` suites already exercise the helper through both public callers, providing regression coverage for the extraction.
+- **Validation:** `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 461 passed (was 460; +1 new). Net −13 LOC across 3 files.
+- **Learning Log:** `[integrated]` — applied the cycle-2 lesson (roam-dev-loop#3): formatting was run scoped to the 3 changed files (`black`/`autoflake` on explicit paths) rather than tree-wide `./format.sh`, so no unrelated drift entered the diff.
+
 ### 2026-06-06 — Fix death-penalty rounding and un-skip silently-dead tests (issues #369, #371)
 - **`src/screen/worldScreen.py`:** Changed the on-death score penalty from `math.ceil(score * 0.9)` to `math.floor(score * 0.9)`. With `ceil`, the intended 10% penalty was a no-op for every score 1–9 (`ceil(0.9)` = 1, `ceil(8.1)` = 9) and only began reducing the score at 10 — exactly the early game where new players die most. `floor` makes the penalty always apply.
 - **`tests/screen/test_worldScreen_deathPenalty.py`** (new): Parametrized regression test pinning `removeEnergyAndCheckForPlayerDeath` at boundary scores (1→0, 9→8, 10→9, 25→22, 100→90); each case would fail under the old `ceil` logic. Also asserts the death counter increments.
--- a/src/config/config.py
+++ b/src/config/config.py
@@ -240,9 +240,7 @@ class Config:
             debug=self.debug,
         )
 
-    def saveWindowSize(self, width, height):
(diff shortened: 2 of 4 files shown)


dmccoystephenson (commented):

- **Scope:** PASS — 4 files, each necessary for #366 or the required CHANGELOG. The `keyBindings.py` logger removal is not unrelated churn: `_logger`/`getLogger` were orphaned by moving the only warning into `Config`. Formatting was run scoped to the 3 changed files, so no unrelated drift entered the diff.
- **Tests-new:** PASS — the new `Config._writeKeyValues` helper has a direct test (`test_write_key_values_updates_appends_and_preserves`, covering update-in-place + append + comment preservation) plus indirect coverage through both public callers.
- **Behavior preservation:** PASS — the extracted loop is byte-for-byte the original; `saveWindowSize` and `saveToConfigFile` build the same `savedValues` and pass their original log message through `errorMessage`. Full suite 461 passed (was 460), including the unchanged `saveWindowSize`/`saveToConfigFile` regression suites.
- **Sibling structure:** PASS — `_writeKeyValues` follows the codebase's protected-method convention (leading underscore + camelCase, as in `_buildRoomFilePath`, `_loadOrGenerateRoom`); the new test mirrors the existing `tmp_path`/`monkeypatch` pattern in `test_config.py`.
- **Issue resolution:** PASS — both named surfaces (`config.py:243`, `keyBindings.py:129`) are changed and the loop is now single-sourced.
- **CI:** PASS — the `test` check is green on the PR head.
- **CHANGELOG updated:** PASS — dated `[integrated]` entry; notes the cycle-2 scoped-formatting lesson was applied.

**One judgment call (deferred, not changed here):** `KeyBindings.saveToConfigFile` calls `config._writeKeyValues(...)`, which reaches a single-underscore (protected) method on a *different* class. This is mildly unidiomatic, but it is the exact API proposed in issue #366 (`Config._writeKeyValues` with `KeyBindings` delegating to it), and `Config` is the natural owner of config-file I/O (it owns `getConfigFilePath`). A public name (e.g. `writeKeyValuesToConfigFile`) would avoid the protected-access smell; kept as specced and flagged for reviewer preference.

Issue #362, opened 2026-05-30
docs: copilot-instructions.md repository layout and Pillow version are stale


dmccoystephenson:
## Summary

`.github/copilot-instructions.md` was found to drift from the current source tree and dependencies during a triage pass.

**Repository Layout** omits three `src/` packages that now exist:
- `src/codex/` — `codex.py`, `codexJsonReaderWriter.py`
- `src/crafting/` — `recipe.py`, `recipeRegistry.py`
- `src/gameLogging/` — `logger.py`

**Technology Stack** lists `Pillow 9.4.0`, but `requirements.txt` pins `Pillow==12.2.0`.

## Suggested change

- Add `codex/`, `crafting/`, and `gameLogging/` to the documented repository layout.
- Correct the Pillow version to match `requirements.txt`.
- Per this file's own learning-log convention, pair the edit with a `CHANGELOG.md` entry.

_Filed by Claude during an automated triage pass; claims above were verified against source._

Issue #346, opened 2026-04-19
Add storage containers


dmccoystephenson:
## Summary

Add craftable storage containers that can be placed in the world, accessed by right-clicking, and persisted between sessions.

## Background

Currently the only way to store items outside the player's inventory is to place them directly in the world, which is lossy and inconvenient. A dedicated storage container entity would give players a proper item management option.

### Entity
- Add a `Chest` entity (`src/entity/chest.py`) — solid, placeable, with a placeholder asset (`assets/images/chest.png`)

### Crafting
- Add a `Chest` recipe to `RecipeRegistry` (`src/crafting/recipeRegistry.py`), e.g. 6× OakWood

### Interaction
- Right-clicking a placed `Chest` in the world opens a chest inventory screen (similar to `InventoryScreen`)
- The chest screen should display the chest's contents and allow items to be moved in/out
- Left-click outside the chest panel drops cursor items (consistent with `InventoryScreen` behaviour)

### Persistence
- Chest contents must survive save/load cycles
- Extend `RoomJsonReaderWriter` (`src/world/roomJsonReaderWriter.py`) to serialize/deserialize `Chest` entities and their contents
- Follow the existing entity registry pattern in `inventoryJsonReaderWriter.py` and `roomJsonReaderWriter.py` when adding the new entity type

### Pickup
- Add `Chest` to the `canBePickedUp()` check in `worldScreen.py` (only when empty, to prevent item loss)

## Acceptance Criteria
- [ ] `Chest` can be crafted and placed in the world
- [ ] Right-clicking a chest opens its inventory UI
- [ ] Items can be moved between the player's inventory and the chest
- [ ] Chest contents are saved to the room JSON and restored on load
- [ ] A non-empty chest cannot be picked up
- [ ] All existing tests pass; new tests cover chest save/load and the canBePickedUp guard

Issue #345, opened 2026-04-19
Prevent floors from being placed on floors


dmccoystephenson:
## Summary

Prevent floor entities from being placed on top of existing floor entities.

## Background

Players can currently stack multiple floor tiles (`WoodFloor`, `StoneFloor`) on the same grid location. This wastes inventory items and is unintended behaviour.

## Requirements

- In `worldScreen.py`, extend the placement validation in `executePlaceAction()` to check whether the target location already contains a floor entity before placing
- Floor types to guard against: `WoodFloor`, `StoneFloor`
- If a floor is already present, silently skip placement (or set a brief status message via `status.set(...)`, consistent with other placement feedback)

## Acceptance Criteria
- [ ] Placing a `WoodFloor` or `StoneFloor` on a location that already contains either floor type is blocked
- [ ] No items are consumed when placement is blocked
- [ ] All existing tests pass; a new test in `tests/screen/test_worldScreen_pushStone.py` or a dedicated placement test covers this guard

Issue #338, opened 2026-04-19
Persist HUD element layout between sessions


dmccoystephenson:
## Summary

Persist HUD element positions between sessions so player layout customizations are preserved across launches.

## Background

`HudDragManager` (`src/ui/hudDragManager.py`) tracks per-element drag offsets at runtime, but these are never written to disk. On every launch the layout resets to defaults regardless of any repositioning the player did in the previous session.

## Requirements

- In `HudDragManager`, add `save()` and `load()` methods that read/write element offsets to `config.yml` using the existing `Config.saveWindowSize()` pattern (key-per-element, e.g. `hudOffset_hotbar_x`, `hudOffset_hotbar_y`)
- Call `hudDragManager.save()` from `Roam.quitApplication()` alongside the existing `config.saveWindowSize()` call
- Call `hudDragManager.load()` during world screen initialization, after HUD elements are registered with the drag manager
- If no saved offsets exist for an element, fall back to its default position (current behaviour)
- Saved offsets must be clamped through the existing `clampPosition()` logic on load to handle resolution changes between sessions

## Acceptance Criteria
- [ ] HUD element offsets are written to `config.yml` on game exit
- [ ] Offsets are restored correctly on the next launch
- [ ] Missing or malformed offset entries fall back to defaults without crashing
- [ ] Layout is still clamped to screen bounds on load (handles resolution changes)
- [ ] All existing tests pass; new tests cover save/load round-trip and missing-key fallback

PR #380, opened 2026-06-07
refactor: consolidate room save-file path into Config.getRoomFilePath

### LINKED ISSUE #373: refactor: room save-file path is built five times across four files while two unused helper methods already exist

dmccoystephenson:
## Summary

DRY refactor giving the room save-file path layout (`<saveDir>/rooms/room_<x>_<y>.json`) a single source of truth.

The path was hand-concatenated in **five** places across four files, with two redundant helpers:
- `src/world/map.py` (`getRoom`), `src/world/roomPreloader.py`, `src/screen/worldScreenPersistence.py` (`saveRoomToFile`) built it inline.
- `src/screen/worldScreen.py` had `_buildRoomFilePath` (used at 2 sites); `worldScreenPersistence.py` had `buildRoomPath` (**zero callers — dead code**).

Changes:
- Added `Config.getRoomFilePath(self, x, y)` as the canonical builder (`Config` already owns `pathToSaveDirectory`).
- Routed all five sites through it; removed both redundant helpers.
- After this change `"/rooms/room_"` appears in exactly **one** place.

## Why

The on-disk room-file layout was encoded in five spots. Any change to it (a directory rename, filename scheme change, or the atomic-write rework in #370) had to be made consistently in all five or the game would read from one location and write to another. This collapses it to one.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 462 passed (was 461; +1 new)
- [x] **Behavior preservation:** new `test_get_room_file_path_format` pins the exact path string (including negative coords); the existing `test_map.py` / `test_roomPreloader.py` load/generate suites exercise the consolidated path end-to-end and pass unchanged.
- [x] `tests/conftest.py`: the shared `test_config` mock now delegates `getRoomFilePath` to the real `Config` implementation, so path construction still reflects each test's `pathToSaveDirectory` (necessary now that the logic sits behind a Config method the `MagicMock(spec=Config)` would otherwise stub).
- [x] Formatting scoped to changed **hunks** — `worldScreen.py` had pre-existing Black drift on unrelated lines, so it was reverted and the two intended edits re-applied by hand to keep the diff focused.

## Note on do-not-auto-merge

`worldScreenPersistence.py` is persistence-adjacent, but the **on-disk format is unchanged** — only how the path string is assembled. The path output is byte-identical (pinned by the new test), so this does not alter save files.

- **#370, #363** (save-corruption / schema-write bugs) — larger; #373 was done first to give them a single path helper to build atomic writes on.
- **#372, #367, #364, #368** (coverage / robustness) — out of this PR's theme.

Closes #373

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,16 @@ logged in detail below.
 
 ## AI Agent Sessions
 
+### 2026-06-06 — Consolidate room save-file path into a single source of truth (issue #373)
+- **`src/config/config.py`:** Added `Config.getRoomFilePath(self, x, y)` — the single canonical builder for the `<saveDir>/rooms/room_<x>_<y>.json` layout.
+- **`src/world/map.py`, `src/world/roomPreloader.py`:** Replaced inline hand-concatenated path builds in `getRoom` / the preloader with `self.config.getRoomFilePath(x, y)`.
+- **`src/screen/worldScreenPersistence.py`:** `saveRoomToFile` now calls `getRoomFilePath`; removed the dead `buildRoomPath` method (it had zero callers).
+- **`src/screen/worldScreen.py`:** Removed the `_buildRoomFilePath` helper and routed its two call sites (`saveRoomToFileAsync`, entity-move save) through `config.getRoomFilePath`.
+- **`tests/conftest.py`:** The shared `test_config` mock (`MagicMock(spec=Config)`) now delegates `getRoomFilePath` to the real `Config.getRoomFilePath` so path construction reflects each test's `pathToSaveDirectory` — required because the logic moved behind a Config method the mock would otherwise stub out.
+- **`tests/config/test_config.py`:** Added `test_get_room_file_path_format` pinning the exact on-disk path string (including negative coordinates) that all save/load call sites depend on.
+- **Validation:** `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 462 passed (was 461; +1 new). The format string `"/rooms/room_"` now appears in exactly one place. Net −22 LOC.
+- **Learning Log:** `[not yet integrated]` — refines the cycle-2 lesson (roam-dev-loop#3): scoping formatters to *changed files* is still insufficient when a touched file carries pre-existing Black drift (`worldScreen.py` here), because `black <file>` reformats the whole file. The fix is to scope to changed *hunks* — after formatting, diff the touched file and revert any reformat that isn't part of the change (here, `worldScreen.py` was reverted and the two intended edits re-applied by hand).
+
 ### 2026-06-06 — Deduplicate the config-file rewrite loop (issue #366)
 - **`src/config/config.py`:** Extracted the read → update-matching-keys → append-new-keys → write-back loop into a new `Config._writeKeyValues(self, savedValues, errorMessage)` helper. `saveWindowSize` now just builds its `savedValues` dict and delegates; the differing log message is passed through `errorMessage` so existing log behavior is preserved.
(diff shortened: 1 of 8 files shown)


dmccoystephenson (commented):

- **Scope:** PASS — 8 files, each necessary for #373 or the required CHANGELOG. `worldScreen.py` carried pre-existing Black drift, so it was reverted and the two intended edits re-applied by hand; the diff shows only the `_buildRoomFilePath` removal and its two call-site replacements.
- **Tests-new:** PASS — `Config.getRoomFilePath` has a direct format test (`test_get_room_file_path_format`, incl. negative coords) plus end-to-end coverage via the `test_map.py` / `test_roomPreloader.py` load/generate suites.
- **Behavior preservation:** PASS — `getRoomFilePath` reproduces the original `str(x)`/`str(y)` concatenation byte-for-byte; the format string `"/rooms/room_"` now appears in exactly one place. Full suite 462 passed (was 461).
- **Sibling structure:** PASS — `getRoomFilePath` mirrors the existing `getConfigFilePath` naming and sits among Config's path/save methods.
- **Issue resolution:** PASS — all five named construction sites now route through the helper; the dead `buildRoomPath` (zero callers) and the duplicate `_buildRoomFilePath` are removed.
- **CI:** PASS — `test` check green on the PR head.
- **Schema/format sync:** PASS — `worldScreenPersistence.py` is touched but the on-disk room-file *format* is unchanged (only path assembly), so no save-file drift and no `schemas/*.json` change is warranted.
- **CHANGELOG updated:** PASS — `[not yet integrated]` entry capturing a refinement to the cycle-2 formatting lesson.

**Two transparency notes (deferred, not changed here):**
1. **Shared test fixture.** `tests/conftest.py`'s `test_config` mock now delegates `getRoomFilePath` to the real `Config` method. This was required because the path logic moved behind a Config method that `MagicMock(spec=Config)` would otherwise stub to a bare mock. It is a change to shared infrastructure; the full 462-test suite confirms no unrelated test depends on `getRoomFilePath` being an inert mock.
2. **Out-of-scope cousin.** `worldScreenPersistence.saveRoomToFile` still builds the `<saveDir>/rooms` *directory* string inline (`pathToSaveDirectory + "/rooms"`) for its `makedirs` call. That is the rooms subdirectory, not the room file path named in #373, so it was intentionally left untouched; it could fold into a future `getRoomsDirectory()` helper if desired.

Issue #239, opened 2025-10-15
Provide text-based UI option


dmccoystephenson:
Users should have the option of using a text-based UI instead of the standard pygame UI.

Issue #234, opened 2025-06-01
Delegate spatial mechanics to Viron server


dmccoystephenson:
A Viron server should be used to store the base associations between environments, grids, locations and entities.

This will require a different approach to data management including saving changes immediately instead of upon shutdown.

Issue #231, opened 2024-07-15
Expose REST endpoint for client extensions


dmccoystephenson:
A REST endpoint should be created to allow client applications to get data about the world for external use in order to extend the functionality of Roam. An example would be a world viewer application or a mapping application.

PR #381, opened 2026-06-07
fix: don't crash the render loop on a missing or unreadable asset

### LINKED ISSUE #368: robustness: a missing or unreadable asset crashes the render loop via DrawableEntity.getImage

dmccoystephenson:
## Summary

Hardens rendering so a single bad sprite degrades gracefully instead of crashing an in-progress session.

`DrawableEntity.getImage` (`src/entity/drawableEntity.py`) called `pygame.image.load` with no error handling. A missing, renamed, or corrupt asset raised `pygame.error` / `FileNotFoundError` straight out of the hot render path (`room.py:87`, `worldScreen.py:1422`/`:1580`, `inventoryScreen.py:141`/`:493`) — a hard crash mid-session, with the image cache ensuring the failure was deterministic per asset.

Changes:
- Wrapped the load in `try`/`except (pygame.error, FileNotFoundError)`.
- On failure: log the offending `imagePath` once (structured `getLogger` warning) and cache a magenta placeholder `pygame.Surface` from the new `_createFallbackSurface()` helper.
- The fallback is cached per path, so the failing load is not retried each frame and the warning is logged exactly once.

## Why a magenta placeholder

It keeps the game running and makes the missing texture obvious in-game (the classic "missing texture" colour) rather than silently invisible or fatal.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 464 passed (was 462; +2 new)
- [x] **Regression evidence:** `test_get_image_returns_placeholder_for_missing_asset` calls `getImage` on a nonexistent path and asserts a fallback `Surface` is returned — this **raises under the old unguarded load**. `test_get_image_caches_fallback_without_retry` asserts the second call returns the same cached surface (no per-frame retry).
- [x] Formatting scoped to the 2 changed files and verified at the hunk level.

- **#370, #363** (corrupt-save crash / invalid-data-written) — the remaining save-durability bugs; larger, need atomic-write design and their own regression tests.
- **#364** (WorldScreenPersistence coverage) — partially stale: it suggests testing `buildRoomPath`, which was removed as dead code in #380; noted on the issue for re-scoping.
- **#372, #367** (Room/Inventory/MapImageGenerator coverage) — out of this PR's theme.

Closes #368

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,12 @@ logged in detail below.
 
 ## AI Agent Sessions
 
+### 2026-06-06 — Harden render loop against missing/unreadable assets (issue #368)
+- **`src/entity/drawableEntity.py`:** `getImage` previously called `pygame.image.load` with no error handling; a missing, renamed, or corrupt sprite raised `pygame.error` / `FileNotFoundError` straight out of the hot render path (`room.py`, `worldScreen.py`, `inventoryScreen.py`), hard-crashing an in-progress, auto-saving session. The load is now wrapped in `try`/`except (pygame.error, FileNotFoundError)`: on failure it logs the offending `imagePath` once (structured `getLogger` warning) and caches a magenta placeholder surface from the new `_createFallbackSurface()` helper, so the game degrades gracefully and the missing asset is visible rather than fatal. The fallback is cached per path, so the failing load is not retried every frame and the warning is logged only once.
+- **`tests/entity/test_drawableEntity.py`:** Added a headless-pygame fixture plus two tests: a missing asset returns a `pygame.Surface` of the fallback size (would raise under the old unguarded load), and the fallback is cached (the second `getImage` returns the same surface, confirming no per-frame retry). Both pop their cache keys to keep the class-level `_imageCache` clean.
+- **Validation:** `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 464 passed (was 462; +2 new).
+- **Learning Log:** `[integrated]` — applied the cycle-4 refinements: formatting was scoped to the two changed files and the diff verified at the hunk level (no pre-existing drift in this file to leak); no `print()`, structured `getLogger` used per logging conventions.
+
 ### 2026-06-06 — Consolidate room save-file path into a single source of truth (issue #373)
 - **`src/config/config.py`:** Added `Config.getRoomFilePath(self, x, y)` — the single canonical builder for the `<saveDir>/rooms/room_<x>_<y>.json` layout.
 - **`src/world/map.py`, `src/world/roomPreloader.py`:** Replaced inline hand-concatenated path builds in `getRoom` / the preloader with `self.config.getRoomFilePath(x, y)`.
--- a/src/entity/drawableEntity.py
+++ b/src/entity/drawableEntity.py
@@ -1,11 +1,15 @@
 import pygame
 from lib.pyenvlib.entity import Entity
+from gameLogging.logger import getLogger
+
+_logger = getLogger(__name__)
 
 
 # @author Daniel McCoy Stephenson
 # @since August 5th, 2022
 class DrawableEntity(Entity):
     _imageCache = {}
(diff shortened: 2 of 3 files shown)


dmccoystephenson (commented):

- **Scope:** PASS — 3 files (`drawableEntity.py`, its test, CHANGELOG), each necessary for #368. No unrelated drift (the file had none to leak under file-scoped Black).
- **Tests-fix (regression):** PASS — `test_get_image_returns_placeholder_for_missing_asset` calls `getImage` on a nonexistent path and asserts a fallback `Surface`; this raises `pygame.error`/`FileNotFoundError` under the old unguarded load, so it fails without the fix.
- **Tests-new:** PASS — the new `_createFallbackSurface` path and the per-path caching are both exercised (`test_get_image_caches_fallback_without_retry` asserts the second call returns the same surface — no per-frame retry/re-log).
- **Fallback size is safe:** PASS — verified all five call sites (`room.py:87`, `worldScreen.py:1422`/`1580`, `inventoryScreen.py:141`/`493`) wrap `getImage()` in `pygame.transform.scale(...)`, so the fixed 32×32 placeholder is rescaled to the target rect and never misaligns; `transform.scale` handles a plain non-alpha surface.
- **Precondition unchanged:** PASS — `_createFallbackSurface` needs pygame initialized, but so did the original `pygame.image.load`, and `getImage` is render-path-only (post-init); no new precondition introduced.
- **Sibling structure:** PASS — the test extends the existing `test_drawableEntity.py` and adds the headless-pygame `init_pygame`/`pygame.quit()` fixture per the project's testing convention.
- **Issue resolution / CI:** PASS — the named surface (`getImage`) is guarded; `test` check green on the PR head.
- **No `@property` / no new `print()` (uses structured `getLogger`) / no `container` import / vendored `src/lib` untouched:** PASS — confirmed via grep over added lines.
- **CHANGELOG updated:** PASS — `[integrated]` entry.

**One judgment call (deferred, not changed here):** the two new tests pop their key from the class-level `DrawableEntity._imageCache` at the start and end, but the trailing `pop` is skipped if an assertion fails mid-test, which would leak a fallback surface into the shared cache for later tests. Impact is low (each test uses a unique `__does_not_exist__` path that no other test references), so a teardown `fixture`/`finally` was judged not worth the extra scaffolding here; flagged for reviewer preference.

PR #382, opened 2026-06-07
fix: abort save on schema-validation failure instead of writing invalid data

### LINKED ISSUE #363: Save: schema validation failures are logged but invalid data is still written to disk

dmccoystephenson:
## Summary

Fixes #363: two JSON writers validated their data against a schema, logged any failure, then **wrote the invalid data anyway** — so validation gave no real protection against persisting a corrupt save.

- `src/inventory/inventoryJsonReaderWriter.py` (`saveInventory`)
- `src/codex/codexJsonReaderWriter.py` (`save`)

Both now `return` on a `ValidationError` (after an explanatory log) instead of falling through to the write. Because validation runs **before** the `open(path, "w")` that truncates the target, aborting leaves the existing good file untouched byte-for-byte.

## ⚠️ Left for maintainer review (not auto-merged)

Two reasons this is handed off rather than auto-merged:
1. These are player-save serializers (`*JsonReaderWriter.py`).
2. **Design tradeoff:** aborting on validation failure converts a *best-effort* save into *strict-or-nothing*. If a schema were ever too strict (a false-positive), a legitimate save would now be **silently dropped on exit** rather than written. This is the behavior #363 asks for, but it is a save-policy change worth a maintainer's explicit sign-off. An alternative policy (temp-file + `os.replace` only after validation passes) is noted under #370.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 468 passed (was 464; +4 new)
- [x] **Regression evidence (verified empirically):** with the production fix temporarily stashed, the three new abort tests **fail** (old code logged the error then re-wrote the file); they pass with the fix. Tests force a `ValidationError` via `monkeypatch` of `jsonschema.validate`, seed a sentinel "good" file, and assert it is neither overwritten nor newly created on failure. A codex save/load round-trip test is also added.
- [x] Formatting scoped to changed files **and** hunks (file-scoped Black collapsed a pre-existing-drift block in the existing `test_saveInventory`; that reformat was reverted so the diff is additions-only apart from the two fix hunks).

## Scope

Deliberately limited to the two writers #363 names. The broader save-durability work — atomic writes (temp + `os.replace`) and guarded `json.load` across all serializers — is **#370**, left as its own follow-up; this PR is the minimal fix for the "invalid data still written" defect.

- **#370** (corrupt-save crash / non-atomic writes) — larger; spans ~13 serializer/loader files; this PR lays the abort-on-invalid groundwork.
- **#372, #367, #364** (coverage) — out of this PR's theme; #364 also partially stale (noted on the issue).

Closes #363

---


dmccoystephenson (commented):

- **Scope:** PASS — 5 files, each necessary for #363 or the required CHANGELOG. File-scoped Black collapsed a pre-existing-drift block in the existing `test_saveInventory`; that reformat was reverted so the test diff is additions-only apart from the two fix hunks.
- **Tests-fix (regression):** PASS — empirically verified: with the production fix temporarily stashed, all three abort tests FAIL (the old code logged the `ValidationError` then re-wrote the file); they pass with the fix.
- **Tests-new:** PASS — `codexJsonReaderWriter` had no test file; added a save/load round-trip plus the abort test. Inventory gained abort + no-create-on-failure tests.
- **Sibling structure:** PASS — the new `tests/codex/test_codexJsonReaderWriter.py` mirrors the inventory writer test's `resolve`/`tmp_path`/`test_config` conventions.
- **CI / no `@property` / no new `print()` (structured `getLogger`) / vendored `src/lib` untouched:** PASS — confirmed via grep over added lines; `test` check green.
- **Schema sync:** PASS — no `schemas/*.json` change and no serialization-format change; the fix is pure control flow (abort before the truncating write), so existing saves remain readable.
- **CHANGELOG updated:** PASS — `[integrated]` entry, including the design tradeoff.

**Two judgment calls (flagged for maintainer review; this PR is intentionally NOT auto-merged):**
1. **Partial policy unification.** #363's *named defect* — `saveInventory` / codex `save` logging a `ValidationError` then writing the invalid data anyway — is fully fixed. But #363's *suggested fix* also asks for "one consistent policy across all JSON writers." The other writers (`stats.py`, `worldScreenPersistence.py`, `tickCounter.py`) do not wrap `validate()` at all, so they `raise` on failure rather than write-anyway. Unifying those is broader, touches more persistence files, and overlaps the atomic-write rework in #370, so it was deliberately left out. The defect this issue describes is resolved; the full cross-writer harmonization is not.
2. **Save-policy tradeoff.** Aborting on validation failure makes the save strict-or-nothing: a false-positive schema rejection would now silently drop a legitimate save on exit instead of persisting it. This is what #363 requests, but it is a behavioral change on player-data serializers — hence the hand-off for explicit maintainer sign-off rather than auto-merge.

dmccoystephenson:
Updated in response to review of this PR. The actionable finding — that an aborted save was silent to the player — was addressed by Claude:

- `saveInventory` / codex `save` now return a success boolean (`False` on abort, `True` on write).
- `WorldScreen.savePlayerInventoryToFile` / `saveCodexToFile` surface a status message ("Could not save inventory/codex (invalid data)") when the writer reports a skipped save, covering both the autosave and on-exit save paths.
- Tests were extended to assert the return contract, and `tests/screen/test_worldScreen_saveSurfacing.py` was added to verify the status message fires on failure and not on success.

The remaining review findings — cross-file save desync (a stale inventory/codex file alongside a freshly written world) and the unguarded `loadInventory` path — are systemic and were left to #370's atomic-write + guarded-load + save-transaction rework, since solving them here would expand well beyond #363's scope. Full suite: 471 passed. Reviewed and updated by Claude.

Issue #412, opened 2026-06-07
robustness: minimap image-load failure is swallowed with no log, hiding a silently broken minimap


dmccoystephenson:
## Summary
In the minimap reload path, a failure to load the map image is caught and silently absorbed with no logging:

- `src/screen/worldScreen.py:1113-1121` —
  ```python
  try:
      mapImage = pygame.image.load(mapImagePath)
      self._cachedMiniMapImage = mapImage
      self._miniMapLastLoadTick = currentTick
  except (FileNotFoundError, pygame.error):
      if self._cachedMiniMapImage is not None:
          mapImage = self._cachedMiniMapImage
      else:
          return
  ```

The fallback-to-cached-or-return is correct (it must not crash the render loop), but the `except` records nothing, so a persistently missing or corrupt `mapImage.png` makes the minimap silently never appear with zero diagnostic.

## Why it matters
This is the same class of silent-render-failure that #368 addressed for `DrawableEntity.getImage` (which now logs once and degrades gracefully). Here the degradation is graceful but undiagnosable — a player reporting "the minimap doesn't show" gives a developer no log to go on.

## Suggested fix
Log once (or throttled) inside the `except`, using the structured logger already imported in this module (`_logger.warning("could not load minimap image", path=mapImagePath, error=str(...))`). Guard against per-frame spam — e.g. only log when transitioning from a previously-good state, since this path runs roughly every 60 ticks. Keep the existing fallback behavior unchanged.

_Filed by Claude during an automated triage pass; claims above were verified against source._

Issue #411, opened 2026-06-07
refactor: JSON schemas are re-read from disk and re-parsed on every save and load in 4 persistence classes


dmccoystephenson:
## Summary
Four persistence classes open and `json.load` their validation schema from disk in **both** `save()` and `load()`, even though the schema file never changes at runtime:

- `src/stats/stats.py:66` (save) and `:83` (load) — `open("schemas/stats.json")`
- `src/world/tickCounter.py:62` (save) and `:76` (load) — `open("schemas/tick.json")`
- `src/goals/goalsJsonReaderWriter.py:22` (save) and `:46` (load) — `open("schemas/goals.json")`
- `src/codex/codexJsonReaderWriter.py:20` (save) and `:49` (load) — `open("schemas/codex.json")`

Each call re-opens and re-parses the schema. Example (`stats.py`):
```python
with open("schemas/stats.json") as f:
    statsSchema = json.load(f)
jsonschema.validate(jsonStats, statsSchema)
```

## Why it matters
Saves happen frequently during play (room transitions, periodic persistence), so this is repeated, avoidable file I/O + JSON parsing on the save path. It's also four copies of the same load-schema boilerplate, each a place a relative-path bug could creep in.

## Suggested fix
Load each schema once and reuse it — e.g. read it in `__init__` (or a module-level cached `_loadSchema(name)` helper), then pass the cached dict to `jsonschema.validate` in both `save()` and `load()`. This removes the duplication and the per-operation disk read.

_Filed by Claude during an automated triage pass; claims above were verified against source._

Issue #393, opened 2026-06-07
Code-sign the Windows installer to avoid SmartScreen warnings


dmccoystephenson:
## Summary
Code-sign `RoamSetup.exe` (and ideally the bundled `Roam.exe`) so Windows SmartScreen / "Unknown publisher" warnings don't scare users away during install.

## Background
#385 (PRs #390, #391, #392) produced an unsigned `RoamSetup.exe` via Inno Setup. Because it's unsigned, Windows shows a SmartScreen "Windows protected your PC / unknown publisher" prompt, and browsers may warn on download. Signing with a trusted certificate removes (or, for a fresh cert, gradually reduces via reputation) these warnings.

## What's needed
- **A code-signing certificate.** Options:
  - Standard OV cert (cheaper, ~\$100–200/yr) — reputation builds over time, so SmartScreen may still warn until enough installs accrue.
  - EV cert (more expensive, hardware/HSM-backed) — immediate SmartScreen reputation.
  - This is a real-world cost/identity decision for the maintainer; it cannot be done purely in code.

## Implementation sketch (once a cert exists)
- Store the cert + password as GitHub Actions secrets (or use a cloud signing service / Azure Trusted Signing).
- Sign in CI with `signtool sign /fd SHA256 /tr <timestamp-url> /td SHA256 ...` — sign `dist\Roam\Roam.exe` before packaging, and sign `installer-output\RoamSetup.exe` after Inno Setup builds it.
- Inno Setup also supports a `SignTool` directive to sign the installer (and uninstaller) during compilation.
- Gate the signing steps on the secret being present so forks/PRs without the cert still build unsigned.

## Acceptance criteria
- Released `RoamSetup.exe` is Authenticode-signed with a valid timestamp; `signtool verify /pa` passes.
- The CI signing steps no-op gracefully when the signing secret is absent.
- README documents the signing setup.

## Notes
Blocked on obtaining a certificate. Until then this stays open. Follow-up to #385.

---
*drafted by Claude on behalf of Daniel Stephenson*

PR #422, opened 2026-06-07
refactor: Single source of truth for the rooms save directory

### LINKED ISSUE #409: refactor: the "/rooms" save-subdirectory creation is duplicated verbatim in two files
## Summary
The check-and-create of the `<saveDir>/rooms` subdirectory is copy-pasted in two places, both rebuilding the path with the raw `self.config.pathToSaveDirectory + "/rooms"` concatenation that #373 set out to centralize (it consolidated the room *file* path into `Config.getRoomFilePath`, but the *directory* creation was left duplicated):

- `src/world/roomJsonReaderWriter.py:83-84` —
  ```python
  if not os.path.exists(self.config.pathToSaveDirectory + "/rooms"):
      os.makedirs(self.config.pathToSaveDirectory + "/rooms")
  ```
- `src/screen/worldScreenPersistence.py:134-135` — the same two lines, verbatim.

## Why it matters
Two copies of the same path literal drift independently: if the rooms layout ever changes (or moves under a new user-data root), one site can be updated and the other missed, writing rooms to the wrong place. It also re-derives a path that `Config.getRoomFilePath(x, y)` already owns.

## Suggested fix
Add a single helper on `Config` (e.g. `getRoomsDirectory()` / `ensureRoomsDirectory()`) that returns/creates `<pathToSaveDirectory>/rooms`, and call it from both save sites — mirroring how `getRoomFilePath` already centralizes the file path (#373).

dmccoystephenson:
## Summary
The `<saveDir>/rooms` path was hand-concatenated in three places — `Config.getRoomFilePath` and the `os.makedirs` in `roomJsonReaderWriter.saveRoom` and `worldScreenPersistence.saveRoomToFile` — so the `"/rooms"` literal could drift independently (the gap #409 names, and a follow-on to #373 which had only consolidated the room *file* path).

- Adds `Config.getRoomsDirectory()` → `<saveDir>/rooms`.
- `getRoomFilePath` is built on top of it; both makedirs sites now call `os.makedirs(self.config.getRoomsDirectory(), exist_ok=True)`.
- **The serialized save format is unchanged** — this only touches where the directory path is built / created, not what is written.

## Test plan
- [x] `pytest` → 536 passed. Added `test_get_rooms_directory_format`; the existing `test_get_room_file_path_format` still pins the same on-disk layout (now produced via `getRoomsDirectory`).
- [x] Delegated `getRoomsDirectory` on the shared `test_config` `MagicMock(spec=Config)` (the spec-mock hazard, roam-dev-loop#4) — without it, `saveRoom` got a MagicMock and 9 room tests failed; verified green after.
- [x] Black-clean on the changed lines (pre-existing drift elsewhere in `test_config.py` left untouched, per the scoped-formatting convention).

Other refactors (#411 schema caching, #412 minimap log) and #364 (persistence tests) left for future cycles; do-not-auto-merge issues (#362/#365/#370) and cert-blocked (#393/#396) for manual review; larger features and install-UX out of scope.

Closes #409

---
*drafted by Claude on behalf of Daniel Stephenson*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ logged in detail below.
 
 | Date | Commits | Summary |
 |------|---------|---------|
+| 2026-06-07 | 1+ | refactor: Add `Config.getRoomsDirectory()` as the single source of truth for the `<saveDir>/rooms` path — `getRoomFilePath` and both makedirs sites (`roomJsonReaderWriter`, `worldScreenPersistence`) now go through it instead of hand-concatenating `"/rooms"` (closes #409) |
 | 2026-06-07 | 1+ | refactor: Replace the 10 near-identical hotbar `elif` branches in `InventoryScreen.handleKeyDownEvent` with a data-driven `_handleHotbarKey` loop (preserving the `hotbar_0 → slot 9` wrap); +3 tests (closes #410) |
 | 2026-06-07 | 1+ | fix: Correct release versioning — `version.txt` had been stale at `0.8.0-SNAPSHOT` while `0.9.0`/`0.10.0` were already released, leading to an erroneous `v0.9.0` tag/release (now deleted); set `version.txt` to `0.11.0-SNAPSHOT`, switch the release workflow to the repo's bare-number tag convention (`0.11.0`, not `v0.11.0`), and fix `UpdateChecker` to read the `/releases` list (every Roam release is a GitHub pre-release, so `/releases/latest` 404s) |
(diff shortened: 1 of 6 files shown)


dmccoystephenson (commented):
Self-review rubric (anchored on green CI: test, windows-installer, package, package-macos):
- Scope: PASS — Config.getRoomsDirectory + its two call sites + getRoomFilePath routing, plus the conftest mock delegation the change requires and a new test.
- Format safety: PASS — only directory-path building/creation changed; generateJsonForRoom and json.dump are untouched, so the on-disk save format is identical (the entry on the do-not-auto-merge list gates on 'whose format changes', which this is not). Room save/load round-trip tests pass.
- Tests: PASS — test_get_rooms_directory_format added; test_get_room_file_path_format still pins the same layout; the spec-mock hazard (roam-dev-loop#4) was hit (9 failures) and fixed by delegating getRoomsDirectory on test_config, then verified green.
- No @property / no print() / DI untouched / vendored lib untouched: PASS.
- Black: PASS on changed lines (pre-existing drift in test_config.py left untouched per the scoped-formatting convention).

Reviewed by Claude on behalf of Daniel Stephenson.

PR #424, opened 2026-06-08
fix: log minimap image-load failures instead of swallowing them

### LINKED ISSUE #412: robustness: minimap image-load failure is swallowed with no log, hiding a silently broken minimap

dmccoystephenson:
## Summary

In `WorldScreen.drawMiniMap`, the minimap reload path caught `(FileNotFoundError, pygame.error)` and silently fell back to the cached frame (or returned). The fallback is correct — it must not crash the render loop — but it logged **nothing**, so a persistently missing or corrupt `mapImage.png` made the minimap silently never appear with zero diagnostic. This is the same class of silent-render-failure that #368 fixed for `DrawableEntity.getImage`.

- The `except` now logs `_logger.warning("could not load minimap image; keeping last good frame if available", path=..., error=...)` **once on the good→failed transition**, guarded by a new `_miniMapLoadFailed` flag.
- A successful load resets the flag, so the warning re-arms if the minimap recovers and later breaks again.
- This avoids per-reload log spam (the path runs roughly every 60 ticks). Existing fallback-to-cached-or-return behavior is unchanged.

## Test plan

New `tests/screen/test_worldScreen_minimap.py` (+3):

- [x] A corrupt `mapImage.png` is logged exactly once and `drawMiniMap` returns cleanly — **fails without the fix** (which logged nothing).
- [x] Repeated load failures log only once (throttle).
- [x] A successful load resets the failure flag (recovery).
- [x] `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` → **539 passed** (was 536).
- [x] Black/autoflake scoped to the two changed files; pre-existing Black drift the file-scoped format pulled in (unrelated `status.set`/`_loadOrGenerateRoom` lines) was reverted so the diff carries only the change.

Closes #412

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ logged in detail below.
 
 | Date | Commits | Summary |
 |------|---------|---------|
+| 2026-06-07 | 1+ | robustness: Log once when the minimap image fails to load — the `except (FileNotFoundError, pygame.error)` in `WorldScreen.drawMiniMap` previously swallowed the failure silently; it now warns via the structured logger on the good->failed transition (throttled by a `_miniMapLoadFailed` flag, reset on a successful load), keeping the existing fallback-to-cached-or-return behavior; +3 tests (closes #412) |
 | 2026-06-07 | 1+ | docs: Sync the config/docs sources of truth — drop the dead `black`/`white` keys from `config.yml` and add the missing `cropGrowthTicks` (static) and `pushableStone` (dynamic toggle) keys the code already reads; correct `copilot-instructions.md` (add the `crafting/`, `codex/`, `gameLogging/` packages to the repository layout, fix the Pillow version to `>=10.0.0`, refresh the stale `0.8.0-SNAPSHOT` version marker to `0.11.0-SNAPSHOT`) (closes #362, #365) |
 | 2026-06-07 | 1+ | refactor: Add `Config.getRoomsDirectory()` as the single source of truth for the `<saveDir>/rooms` path — `getRoomFilePath` and both makedirs sites (`roomJsonReaderWriter`, `worldScreenPersistence`) now go through it instead of hand-concatenating `"/rooms"` (closes #409) |
 | 2026-06-07 | 1+ | refactor: Replace the 10 near-identical hotbar `elif` branches in `InventoryScreen.handleKeyDownEvent` with a data-driven `_handleHotbarKey` loop (preserving the `hotbar_0 → slot 9` wrap); +3 tests (closes #410) |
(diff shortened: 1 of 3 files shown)


dmccoystephenson (commented):

- **Scope:** PASS — only `src/screen/worldScreen.py`, the new test file, and `CHANGELOG.md`; the file-scoped Black drift it initially pulled in was reverted so the diff carries only the intended hunks.
- **Tests-fix:** PASS — `test_corrupt_minimap_image_is_logged_and_does_not_crash` asserts `_logger.warning` is called once; this fails against the pre-fix code, which logged nothing.
- **Tests-new:** PASS — the throttle and recovery branches each have a dedicated test.
- **Sibling structure:** PASS — the test builds `WorldScreen` via `__new__` with minimal attributes, mirroring `tests/screen/test_worldScreen_saveSurfacing.py`.
- **Issue resolution:** PASS — #412's named surface area (the silent `except` in `drawMiniMap`) is the exact code changed.
- **No `@property`:** PASS — grep of added lines returns 0.
- **No new `print()`:** PASS — grep of added `src/` lines returns 0; logging uses the existing module `_logger` (`getLogger`).
- **DI discipline:** N/A — no class construction changed; uses the existing injected `tickCounter`/`config`.
- **CHANGELOG updated:** PASS — dated row + `[integrated]` AI Agent Sessions entry.
- **Vendored lib untouched:** PASS — no `src/lib/` file changed.

Summary: the change was reviewed by Claude; it restores diagnosability for a previously-silent failure while preserving the non-crashing fallback. CI is the green external anchor.

PR #425, opened 2026-06-08
refactor: cache JSON validation schemas instead of re-reading per save/load

### LINKED ISSUE #411: refactor: JSON schemas are re-read from disk and re-parsed on every save and load in 4 persistence classes

dmccoystephenson:
## Summary

Four persistence classes opened and `json.load`-ed their validation schema from disk in **both** `save()` and `load()`, even though schema files never change at runtime — repeated, avoidable disk I/O + JSON parsing on the frequently-hit save path, plus four copies of the same load-schema boilerplate.

- **New `src/schemaCache.py`** — `loadSchema(filename)`, `functools.lru_cache`d: reads `schemas/<filename>` once and returns the cached parsed dict. Uses the same working-directory-relative path the call sites already used (set at startup by `appPaths.prepareWorkingDirectory`), so behavior is unchanged when frozen. Implemented as a module-level cached helper (mirroring the `gameLogging.getLogger` pattern) rather than a DI component, since it has no injectable dependencies and is pure read-only.
- **Routed through it:** `stats.py`, `world/tickCounter.py`, `goals/goalsJsonReaderWriter.py`, `codex/codexJsonReaderWriter.py` — each `with open(...) ... json.load ... jsonschema.validate` pair collapsed to `jsonschema.validate(data, loadSchema("X.json"))` in both methods.

**No serialized format changes** and validation semantics are identical (same schema object passed to `jsonschema.validate`).

## Scope note

The same re-read pattern also exists in `worldScreenPersistence.py`, `roam.py`, `inventoryJsonReaderWriter.py`, and `roomJsonReaderWriter.py`. #411 scopes to the four classes above, so the others are left for a follow-up to keep this diff matched to the issue (the latter two are also on the do-not-auto-merge persistence list).

## Test plan

New `tests/test_schemaCache.py` (+3):

- [x] `loadSchema` returns the parsed schema dict.
- [x] Returns the same cached object on repeat calls.
- [x] Reads each file from disk only once (patched `builtins.open` call count).
- [x] `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` → **542 passed** (was 539; +3).
- [x] Black/autoflake scoped to the six changed files (no drift).

Closes #411


dmccoystephenson (commented):

- **Scope:** PASS — new `schemaCache.py`, the four named persistence classes, a new test file, and `CHANGELOG.md`; no unrelated churn (Black left all six code files unchanged).
- **Tests-new:** PASS — `loadSchema` has tests for parse, caching-identity, and read-once behaviour.
- **Behaviour preserved:** PASS — each site still calls `jsonschema.validate(data, <schema>)`; only the schema's source changed from a fresh `open`+`json.load` to the cached dict. No serialized format changed; the existing per-class save/load tests (`tests/stats`, `tests/world`, `tests/goals`, `tests/codex`) stay green.
- **DI discipline:** PASS — `loadSchema` is a module-level cached function with no injectable dependencies, mirroring the established `gameLogging.getLogger` pattern; intentionally not a `@component` (nothing to wire). No `container` import added to any class file (grep: 0).
- **No `@property` / no new `print()` / vendored lib untouched:** PASS — greps return 0.
- **Issue resolution:** PASS — all four classes #411 names are routed through the cache; the four out-of-scope sites are called out in the PR body as a deliberate follow-up, not silently skipped.
- **CHANGELOG updated:** PASS — dated row + `[integrated]` AI Agent Sessions entry.
- **Schema sync:** N/A — no `JsonReaderWriter` *serialization* changed (schema files and serialized output are byte-identical).

**Judgment call (flagged):** `loadSchema` caches the parsed dict process-wide for the lifetime of the run. This is safe because schema files are static at runtime and `jsonschema.validate` does not mutate the schema; no call site mutates the returned dict. If a future caller needed a hot-reloadable schema, it would have to bypass the cache.

Summary: reviewed by Claude — a behaviour-preserving de-duplication that removes per-save schema I/O. CI is the green external anchor.

PR #426, opened 2026-06-08
fix: tolerate corrupt/truncated save files on load (#370, load half)

dmccoystephenson:
## Summary

Addresses the **load-resilience half** of #370. The startup loaders called `json.load` with no error handling, so a truncated or corrupt save file (e.g. from a power loss / force-quit mid-save) raised `json.JSONDecodeError` straight out of `WorldScreen.initialize` — crashing the game on **every** launch, with no in-game recovery (the file still exists, so the `os.path.exists` guards keep re-attempting the load). The user had to manually delete the file.

- **New `src/jsonPersistence.py`** — `readJsonFile(path, default=None)`: returns the parsed object, returns `default` on a missing file, and logs the path + returns `default` on `(json.JSONDecodeError, OSError, UnicodeDecodeError)`. Module-level helper, mirroring `gameLogging.getLogger` / `schemaCache.loadSchema`.
- **Loaders routed through it:** `stats.load`, `tickCounter.load`, `worldScreenPersistence.loadPlayerLocationFromFile` / `loadPlayerAttributesFromFile`, `roomJsonReaderWriter.loadRoom` (now returns `None` when unreadable), `inventoryJsonReaderWriter.loadInventory`. A damaged file is treated as "no save" and the loader falls back to its existing default.
- **`map.getRoom`** now treats an unreadable room file as absent (returns `-1`), so `initialize` generates a fresh room (0,0) instead of crashing.

**No serialized format change** and schema *validation* behavior (the separate #363 concern) is unchanged.

## Scope note

#370 also calls for **atomic writes** (write-to-temp + `os.replace`) across the writers, to *prevent* corruption in the first place. Together that would touch >10 files, over the dev-loop scope ceiling, so it's split: this PR fixes the crash-on-load symptom; the atomic-write half is a follow-up. **#370 stays open** until that lands — this PR says `Addresses`, not `Closes`.

## Test plan

New `tests/test_jsonPersistence.py` (+7):

- [x] `readJsonFile` — valid / missing / corrupt / truncated.
- [x] `Stats.load` and `TickCounter.load` no longer raise on a corrupt file (**would crash before the fix**) and leave state at defaults.
- [x] A valid file still round-trips (happy path preserved).
- [x] `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` → **549 passed** (was 542; +7).
- [x] Black/autoflake scoped to the eight changed files (no drift).

## Note for reviewer

This touches persistence serializers (`roomJsonReaderWriter.py`, `inventoryJsonReaderWriter.py`, `worldScreenPersistence.py`). It does **not** change the on-disk format — only how unreadable files are handled on load — but flagging it since persistence is data-critical.

Addresses #370

---


dmccoystephenson (commented):

- **Scope:** PASS — new `jsonPersistence.py`, six loader sites, `map.getRoom`, a new test file, and `CHANGELOG.md`; Black left all eight code files unchanged (no drift leaked).
- **Tests-fix:** PASS — `test_stats_load_tolerates_corrupt_file` and `test_tickCounter_load_tolerates_corrupt_file` exercise the exact crash path and would raise `json.JSONDecodeError` against the pre-fix code.
- **Tests-new:** PASS — `readJsonFile` has valid/missing/corrupt/truncated tests; a happy-path round-trip guards against the tolerant read masking real loads.
- **Behaviour preserved:** PASS — schema validation and serialized format are untouched; only the *source* of a load (and the missing/corrupt branch) changed. The `os.path.exists` guards were folded into `readJsonFile` (missing -> default), so semantics for a missing file are identical.
- **Issue resolution:** PASS — every unguarded startup loader #370 names is routed through the tolerant reader, and `map.getRoom` handles the corrupt-room path that crashed via `map.py`. The PR says `Addresses` (not `Closes`) #370 because the atomic-write half is a deliberate, documented follow-up.
- **DI discipline:** PASS — `readJsonFile` is a module-level helper with no injectable deps (mirrors `getLogger`/`loadSchema`); no `container` import added.
- **No `@property` / no new `print()` / vendored lib untouched:** PASS — greps return 0; logging via the module `_logger`.
- **CHANGELOG updated:** PASS — dated row + `[integrated]` AI Agent Sessions entry; `import os` removed from `stats.py` by autoflake (now unused).

**Judgment call (flagged):** A corrupt room (0,0) file is treated as absent, so a fresh room is generated and the next save overwrites the corrupt file — that room's prior contents are lost. This is the intended degradation per #370 (the file was unrecoverable and otherwise crashed every launch); the rest of the save (stats, inventory, other rooms) survives. Flagging because it touches data-critical persistence.

Summary: reviewed by Claude — defensive, format-preserving load hardening that turns a fatal startup crash into graceful degradation. CI is the green external anchor.

PR #427, opened 2026-06-08
fix: make all save writes atomic (#370, write half — closes it)

### LINKED ISSUE #370: bug: a corrupt or truncated save file crashes the game on load — json.load is unguarded and writes are non-atomic

dmccoystephenson:
## Summary

The **write-atomicity half** of #370 — completing the fix (the load-resilience half landed in #426). Every JSON writer opened the existing good file with `open(path, "w")`, which truncates immediately, then `json.dump`ed into that handle. A process kill, full disk, or `json.dump` error mid-write left the good file already truncated and a partial result on disk — permanently breaking the save slot.

- **`writeJsonAtomically(path, data, indent=4)`** added to `src/jsonPersistence.py`: `os.makedirs` the parent dir, write to a `tempfile.mkstemp` temp file **in the same directory**, `flush()` + `os.fsync()`, then `os.replace()` over the target (atomic on a single filesystem). On any exception the temp file is removed and the error re-raised, so a failed save leaves the previous file intact with no stray temp.
- **Every writer routed through it:** `stats.save`, `tickCounter.save`, `codexJsonReaderWriter.save`, `goalsJsonReaderWriter.save`, `inventoryJsonReaderWriter.saveInventory`, `roomJsonReaderWriter.saveRoom`, `worldScreenPersistence` (location / attributes / room), and `WorldScreen._writeJsonToFile` / `_doSave` (the in-game background room-save path).

**Serialized output is byte-identical** (same `json.dump(..., indent=4)`); only the write mechanism changed.

## Scope note

`WorldScreen._writeJsonToFile` / `_doSave` weren't in #370's enumerated writer list, but they're the hot in-game save path with the same truncate-then-dump bug, so they're included — otherwise #370 wouldn't actually be closed.

## Test plan

`tests/test_jsonPersistence.py` (+4):

- [x] Atomic write round-trips; creates a missing parent directory.
- [x] No temp file left after a successful write.
- [x] **A serialization failure preserves the previous good file and leaves no temp file** (the core #370 guarantee).
- [x] `python3 -m compileall src -q` clean; `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` → **553 passed** (was 549; +4).
- [x] Black/autoflake scoped to the ten changed files; `import json` removed where now unused; pre-existing `worldScreen.py` Black drift reverted so the diff carries only the change.

## Note for reviewer

Touches all persistence serializers. On-disk format is unchanged — only writes are made crash-safe. Combined with #426, this fully resolves #370.

Closes #370

---


dmccoystephenson (commented):

- **Scope:** PASS — `writeJsonAtomically` added to `jsonPersistence.py`, every writer routed through it, +4 tests, CHANGELOG. Pre-existing `worldScreen.py` Black drift was reverted so the diff carries only the atomic-write change.
- **Tests-fix:** PASS — `test_writeJsonAtomically_preserves_good_file_when_serialization_fails` is the regression for #370's write half: it forces a mid-write failure and asserts the previous file survives intact with no temp leftover. This fails against the old `open(path,"w")` + `json.dump` pattern (which truncates first).
- **Tests-new:** PASS — round-trip, missing-directory creation, and no-temp-on-success are covered.
- **Behaviour preserved:** PASS — serialized output is byte-identical (`json.dump(..., indent=4)`); only the write mechanism changed. The full per-class save/load round-trip suite stays green (553 passed).
- **Cross-platform:** PASS — the Windows CI job (`windows-installer`) runs the suite, so `os.replace`-based atomic writes are exercised on Windows, not just POSIX (`os.replace` replaces existing files on Windows, unlike `os.rename`).
- **Issue resolution:** PASS — every writer #370 names plus the in-game `WorldScreen._writeJsonToFile`/`_doSave` hot path are converted; `Closes #370` is justified now that both halves (load #426 + write here) are in.
- **DI discipline / no `@property` / no new `print()` / vendored lib untouched:** PASS — module-level helper, greps return 0.
- **CHANGELOG updated:** PASS — dated row + `[integrated]` entry.
- **CI:** PASS — all four jobs green on the head commit.

**Judgment call (flagged):** `writeJsonAtomically` adds an `fsync` per save. Every caller is on a background save thread (`_saveExecutor` / `_doSave`) or an explicit save action, not the render loop, so the extra durability cost does not block the frame loop. Flagging because fsync is the deliberate trade-off for crash-safety.

Summary: reviewed by Claude — completes #370 by making writes crash-safe without changing the on-disk format. CI (incl. Windows) is the green external anchor.

Issue #463, opened 2026-06-14
Complete pygame removal from game logic — residual output/lifecycle/value-type usage (follow-up to epic #433)


dmccoystephenson:
Follow-up to epic #433. The `Renderer` + `InputSource` seam is built and **proven** (Phase 5 `NullRenderer`/`NullInputSource`, #439). All **input** in `src/` goes through `InputSource`. But the epic's aspirational end-state — *"game logic should never import pygame"* — still has residual pygame, all of it **non-input**. Grouped by the work it belongs to:

## Output rendering not yet routed through `Renderer`
- `src/screen/worldScreen.py` — `Surface`/`SRCALPHA`/`BLEND_RGBA_*`/`image` for the minimap and day/night overlays.
- `src/world/dayNightCycle.py` — `Surface`/`SRCALPHA` light masks.
- `src/screen/codexScreen.py` — `image`/`transform`/`font` icon+font cache.
- `src/screen/screenshotHelper.py` — `Surface`/`image` (the screenshot writer).

These were **deliberately deferred** during the output-seam phases (per-pixel alpha compositing, font handles, and the codex's None-for-undiscovered semantics didn't map cleanly onto the minimal `Renderer` API). Completing them means extending `Renderer` with the few needed primitives (offscreen alpha surfaces / blend modes / text-measure) and migrating these call sites.

## Lifecycle
- `src/screen/{mainMenuScreen,statsScreen,inventoryScreen}.py` — `pygame.quit()` in `quitApplication()`. Should route through the frontend (`frontend.quit()`).
- `src/roam.py` — `pygame.init`/`pygame.image` at the composition root (arguably acceptable there, but could move fully into `PygameFrontend`).

## Value-type boundary / display query
- `src/ui/geometry.py` — converts the backend-neutral `Rect` to `pygame.Rect` at the draw boundary (Phase 2a left this as the conversion point).
- `src/config/config.py` — `pygame.display.Info()` to pick a default window size.

## Why this is a follow-up, not a reopen of #433
The epic's deliverable was **the seam** (two interfaces + a pygame implementation + an executable proof), and that is done and merged across Phases 0–5. The remaining items are an incremental "finish the output seam + lifecycle cleanup" pass that can land independently without touching the now-stable input path. Suggest sizing each group as its own PR.

_Filed by Claude on behalf of Daniel Stephenson._

PR #469, opened 2026-06-14
refactor: move the OS display query out of Config (#463)

dmccoystephenson:
## Summary

Next slice of #463. `Config.__init__` called `pygame.display.Info()` directly to default the window to 90% of screen height — pulling pygame into a backend-neutral module.

- New **`rendering/displayInfo.getScreenSize()`** wraps the pygame display query (and the idempotent `display.init()` that handles Config running before the frontend builds the window). pygame stays in `rendering/`, where the grep gate allows it.
- `config.py` calls the helper and **no longer imports pygame**.
- **Behavior preserved exactly** — same query; now computed once up front instead of twice (the redundant `current_w` query inside the saved-size branch is removed).

Note: `ui/geometry.py` and `roam.py` (its `--selftest` deliberately exercises the pygame bundle) — both previously listed under #463 — are *not* real targets: geometry's pygame mention is a comment, and roam's is the bundle self-test. Updated my mental model of #463's remaining surface accordingly (codex image/font and the worldScreen/dayNightCycle overlays remain, and those need `Renderer` API extensions + a visual smoke-run).

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 761 passing (was 760)
- [x] `python3 -m black --check` — clean
- [x] `python3 src/roam.py --selftest` — OK (constructs `Config()`, which now goes through the helper)
- [x] `grep pygame src/config/config.py` — comment only; no import

Part of #463 (codex + overlays to follow as separate PRs; those need a smoke-run).


dmccoystephenson:
Self-review rubric (reviewed by Claude): Scope PASS (new helper + config edit + test + CHANGELOG, all necessary); Behavior-preserving PASS (same screen query, selftest constructs Config OK); Tests-new PASS (getScreenSize returns positive w/h); DI/no-print/no-@property PASS; CHANGELOG PASS; CI PASS (all green incl. black gate). config.py is now pygame-free (comment only); config.yml (the do-not-auto-merge path) is untouched.

---

PR #506, opened 2026-06-28
fix: extract Dockerfile inline Python to web/build_zip.py

dmccoystephenson:
Docker parses multi-line \`RUN python3 -c "..."\` heredocs as Dockerfile instructions, treating `import` as an unknown instruction. Move the zip-building logic to `web/build_zip.py` and call it with `RUN python3 web/build_zip.py`.

---


dmccoystephenson:
Closing in favor of #507 which is rebased on current main

PR #525, opened 2026-07-19
fix: wild living entities can migrate rooms from a corner tile

### LINKED ISSUE #521: Wild living entities can never migrate rooms from a corner tile (player-only corner logic not extended to wildlife)
## Summary

The player's cross-room movement already handles the four corner tiles of a room, but the equivalent path for ordinary living entities (Chicken/Bear/Rabbit/Deer/Wolf/Snake wandering between rooms) does not, so a wild entity that ends up on a corner tile simply stops migrating rooms — silently, forever.

- `src/screen/worldScreen.py:261-300` (`getCoordinatesForNewRoomBasedOnPlayerLocationAndDirection`) — the player's corner case is fully implemented: `if self.isCorner(location): ... direction = self.player.getDirection() ...` with explicit x/y adjustment for all four corners.
- `src/screen/worldScreen.py:302-309` (`getCoordinatesForNewRoomBasedOnLivingEntityLocation`) — the living-entity equivalent instead does `if self.isCorner(location): raise Exception("corner movement not implemented yet")`.
- `src/screen/worldScreen.py:2392-2408` (`_updateLivingEntities`) — the caller wraps this in `try/except Exception` and just does `continue` (only logging in debug mode), so the failure is invisible in normal play; the entity is left in the old room instead of migrating.
- This is the plain-wildlife path (`Room.moveLivingEntities` → `_updateLivingEntities`), distinct from the new NPC room-crossing path recently added in the in-flight `feature/npc-cpc-agents` PR, which handles only NPCs and explicitly skips corners for them too — the underlying wildlife-corner case addressed here is untouched by that PR.

## Why it matters

Any wild animal whose room-exit tile happens to be a grid corner is permanently stuck oscillating within its current room — it can never roam into an adjacent room from that position, which is inconsistent with the player experiencing full room-to-room movement including corners, and silently caps how far wildlife can spread across the generated world.

## Suggested fix

Extract the corner x/y-adjustment logic already written for the player (`src/screen/worldScreen.py:266-290`) into a shared helper parameterized by `direction`, and reuse it in `getCoordinatesForNewRoomBasedOnLivingEntityLocation`. For a living entity (which, unlike the player, has no meaningful "facing direction" input at rest) the equivalent direction can be derived from which edge of the corner cell the entity is adjacent to, or simply picked deterministically (e.g. prefer the horizontal neighbor) — either approach removes the `raise` and lets wildlife cross room boundaries from any tile, matching the player's existing behavior.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

dmccoystephenson:
## Summary

- Wild living entities (Chicken/Bear/Rabbit/Deer/Wolf/Snake) that wandered onto a grid corner tile could never migrate rooms from there again — `getCoordinatesForNewRoomBasedOnLivingEntityLocation` and `getNewLocationCoordinatesForLivingEntityBasedOnLocation` (`worldScreen.py`) both raised on a corner, and `_updateLivingEntities` silently swallowed the exception (`continue`).
- Since living entities have no facing direction (unlike the player, which resolves corners via `player.getDirection()`), corners now simply fall through to the same X-axis-first branches already used for plain edges. This resolves to a consistent horizontal-preference migration and requires no new branching logic — just removing the two `raise` guards.
- Added a regression test module covering all four corners for both helper functions, plus an integration test through `_updateLivingEntities` confirming an entity actually crosses into the adjacent room and is registered there.

Closes #521

## Test plan
- [x] `python3 -m black --check src tests`
- [x] `python3 -m pytest tests/` — 988 passed
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` — 236 passed
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` — 281 passed

---


dmccoystephenson:
test comment

---

dmccoystephenson:
Self-review (posted as a comment since PR review submission required approval unavailable in this session): Verified the fix by tracing both corner-resolution functions symbolically for all 4 corners against the non-corner edge-priority order (X-axis checked before Y-axis) - the result is a consistent horizontal-preference migration matching the player's corner logic in direction, with no new branching needed. Found and fixed one issue (unused Grid import in the new test file, pushed as a follow-up commit) before finalizing. No remaining issues found; full suite (988 tests) and both frontend subsets pass, black is clean.

---

PR #527, opened 2026-07-19
feat: persist HUD element layout between sessions

### LINKED ISSUE #338: Persist HUD element layout between sessions

dmccoystephenson:
## Summary

- `HudDragManager` tracked per-element drag offsets (hotbar/status/energy-bar/minimap) at runtime, but never wrote them to disk — every launch reset the layout to defaults regardless of prior repositioning.
- Added `HudDragManager.save(config)` / `HudDragManager.load(configValues, screenWidth, screenHeight)`, following the existing `Config._writeKeyValues` / `KeyBindings.saveToConfigFile` key-per-value pattern (`hudOffset_<name>_x` / `hudOffset_<name>_y`).
- `Roam.quitApplication()` and the return-to-main-menu branch of `Roam.run()` both save the layout alongside the existing window-size save.
- `WorldScreen.initialize()` calls `hudDragManager.load()` right after HUD elements are registered, using the current display size.
- Missing or malformed offset entries fall back to the default position without crashing; restored offsets are re-clamped through the existing `clampPosition()` logic so a resolution change between sessions can't push an element off-screen.
- **Frontends:** loading applies to both — `_drawTextMinimap()` and the pygame HUD paths both read `hudDragManager.getOffset()`. Saving is gated on `supportsImageLoading()` (inside the same guard the window-size save already uses), because dragging is mouse-only: a `--text` session can never legitimately change the layout, but `load()` clamps offsets to the far smaller terminal-equivalent display, so writing them back would silently shrink a layout arranged in pygame.

Closes #338

## Test plan

- [x] `python3 -m pytest tests/` — 1015 tests pass (+12 new)
- [x] `python3 -m black --check src tests` — no formatting violations
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` — 236 text-interface unit tests pass
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` — 300 pygame-facing screen tests pass

---


dmccoystephenson:
Self-review: implementation matches issue #338's spec closely — `HudDragManager.save()`/`load()` follow the existing `Config._writeKeyValues`/`KeyBindings.saveToConfigFile` key-per-value pattern, wired into `Roam.quitApplication()` and `WorldScreen.initialize()` exactly where the issue specified. Missing/malformed offset pairs fall back to defaults without crashing (verified with a dedicated test), and restored offsets are re-clamped via the existing `clampPosition()` on load. No frontend-specific branching was needed since HUD drag offsets already apply uniformly to both frontends elsewhere in `worldScreen.py` (dragging itself is mouse-only and was already unguarded). Full suite (1007 tests, +8 new) and both text/pygame subsets pass; `black --check` is clean. One pre-existing, unrelated flake: `tests/integration/test_text_playthrough.py` (pty/real-terminal timing) intermittently times out under this sandbox's resource constraints — reproduced the same failures on a clean `main` checkout before making any changes, so it isn't caused by this PR.

---

dmccoystephenson:
@copilot address comments

---

dmccoystephenson:
Follow-up verification after this PR sat open: the branch was behind `main` (PR #528 had merged), so it was rebased onto the current `main` and re-pushed. Re-ran the full verification pass on the rebased head:

- `python3 -m pytest tests/` — 1011 passed
- `python3 -m black --check src tests` — clean
- CI on the rebased head (test, package, package-macos, windows-installer, Save persistence (Playwright)) — all green: https://github.com/Preponderous-Software/roam/actions/runs/29718797243
- Diff re-checked against issue #338's acceptance criteria and the do-not-auto-merge path list — no protected paths touched (CHANGELOG.md, src/roam.py, src/screen/worldScreen.py, src/ui/hudDragManager.py, and two test files only)
- Config value round-trip (`Config.parseConfigValue` coercing saved string values back to int/float on read) matches what `HudDragManager.load()`'s `isinstance(rawX, (int, float))` check expects — confirmed by reading `src/config/config.py`

This PR was left open at the end of this run pending human merge approval (this session was not authorized to merge).

---

dmccoystephenson:
Self-review (continuation run — this PR was left mid-flight by an interrupted dispatch; picked it up rather than opening a duplicate).

**Re-verified the existing work on the current head:**
- `HudDragManager.save()` calling `config._writeKeyValues` is an established cross-class pattern, not a new leak — `src/config/keyBindings.py:163` already does exactly this.
- Round-trip is sound: `save()` writes `str(int(offset))`, and `Config.parseConfigValue` (`src/config/config.py:119`) coerces digit strings back to `int`, which is what `load()`'s `isinstance(raw, (int, float))` guard expects. The explicit `isinstance(raw, bool)` rejection correctly stops `hudOffset_x: true` from being read as `1`.
- `self.worldScreen` is resolved in `Roam.initializeScreens()` before any quit path can run, so the new call sites can't hit an unset attribute; quitting straight from the main menu just writes an empty value set (`_writeKeyValues({})` rewrites the file unchanged), leaving previously saved offsets intact.
- `WorldScreen.initialize()`'s new `self.renderer.getDisplaySize()` is safe in text mode — `TextRenderer.getDisplaySize()` (`src/rendering/textRenderer.py:132`) returns the pixel-equivalent of the terminal grid.

**One real defect found and fixed in `d5f9ca2`:** the save was unguarded, but drag input is mouse-only and `load()` clamps restored offsets to the current display. A single `--text` session would therefore read a pygame layout, clamp the offsets down to the terminal-equivalent display, and write the shrunken values back on quit — silently destroying the desktop layout. Saving is now inside the same `supportsImageLoading()` guard the window-size save already uses, in both `quitApplication()` and the return-to-main-menu branch of `run()`. Loading stays unconditional, since `_drawTextMinimap()` reads `getOffset()` and the TUI should still show the saved layout. Four new tests in `tests/test_roam.py` cover both paths in both modes.

Also corrected the CHANGELOG row, which credited only `quitApplication()` and carried a pre-rebase test count.

Verification on `d5f9ca2`: `pytest tests/` 1015 passed; `tests/rendering/ tests/config/` 236 passed; `tests/screen/ tests/ui/` 300 passed; `black --check src tests` clean.

Note: Copilot's second review left one inline comment that could not be read in this headless run (the `gh api` call needed to fetch inline review comments isn't available to this session), so it has not been explicitly addressed — worth a glance before merge. This session was not authorized to merge, so the PR is left open for human review.

---

dmccoystephenson:
Phase 7 documentation pass on `be6abd7`:

- **README** — the controls table never listed middle-mouse HUD dragging (`WorldScreen.handleMouseDownEvent` → `hudDragManager.handleMouseDown` on `MIDDLE_MOUSE_BUTTON`), which is the only way to produce the layout this PR persists. Added a row for it.
- **PLANNING.md** — Room Types listed only Empty/Grassland/Forest/Jungle/Mountain, omitting `RoomType.CAVE`, which `roomFactory` genuinely generates below ground level. Added it. Re-checked the Mobs and Crafting lists programmatically against `LIVING_ENTITY_TYPES` and `RecipeRegistry` — both already match (6 mobs, 9 recipes).
- **CLAUDE.md** — frontend-abstraction and keybinding contracts unchanged by this PR; `ACTION_LABELS` covers all 36 `DEFAULT_BINDINGS` entries with no gaps.
- **CHANGELOG** — row now names both save paths, the text-mode gate, and the current 1015-test count.

CI is green on `be6abd7` (test, package, package-macos, windows-installer, Save persistence) and GitHub reports the branch `MERGEABLE`/`CLEAN`. Left open for human review and merge — this session had no merge authorization.

---

PR #539, opened 2026-07-29
fix: make three advertised keybindings actually reachable

### LINKED ISSUE #536: Controls screen: the advertised "S: save" key is unreachable, so keybindings can never be saved without a mouse
In `ControlsScreen.handleKeyDownEvent` (`src/screen/controlsScreen.py`), the branch that saves pending rebindings is dead code:

```python
elif key in (KeyCode.DOWN, KeyCode.S):      # line 243 — cursor down
    ...
elif key in (KeyCode.RETURN, KeyCode.KP_ENTER, KeyCode.SPACE):
    ...
elif key == KeyCode.S:                       # line 250 — save; never reached
    if not self.keyBindings.getConflictsForBindings(...):
        self.saveAndReturn()
```

`KeyCode.S` is consumed by the earlier `(KeyCode.DOWN, KeyCode.S)` navigation branch, so pressing `S` just moves the cursor down; the save branch below it can never execute.

This matters for two reasons:

1. **The screen advertises the key.** `drawBindings()` draws the hint line `"Up/Down: navigate  -  Enter/Space: remap  -  S: save  -  R: reset  -  Esc: cancel"` (`src/screen/controlsScreen.py:~137`). The hint documents `Up/Down` for navigation — not `W/S` — so the `W`/`S` aliases are the later addition that broke the documented contract.
2. **Text mode has no fallback.** Saving otherwise requires clicking the on-screen "Save" button, and `TextInputSource.getMousePosition()`/`getMouseButtons()` return `(0, 0)`/`(False, False, False)` unconditionally (`src/rendering/textInputSource.py`), so a `--text` player can remap a binding but has no way at all to persist it — `Esc` runs `cancelAndReturn()` and throws the pending edits away.

Suggested fix: drop `KeyCode.W`/`KeyCode.S` from the two navigation branches on this screen (the arrow keys already cover navigation in both frontends — `TextInputSource` decodes the CSI/SS3 arrow sequences), so `S` saves as the hint says. `R` (reset) has no such conflict and already works.

Add regression coverage that `handleKeyDownEvent(KeyCode.S)` on a conflict-free pending set calls `saveAndReturn()` and leaves the cursor where it was.

### LINKED ISSUE #537: Screenshot is unreachable in --text mode: bound only to Print Screen, which a terminal can never send
The `screenshot` action is bound to `KeyCode.PRINTSCREEN` in `DEFAULT_BINDINGS` (`src/config/keyBindings.py:44`) with no `alt_screenshot` ASCII binding.

`TextInputSource.pollEvents()` (`src/rendering/textInputSource.py`) can only ever emit key codes from three sources: `fromInt(ord(char))` for a typed character, the four arrow escape sequences in `_ARROW_SEQUENCES`, and the `_CONTROL_CHARS` map (Enter/Backspace). `KeyCode.PRINTSCREEN` is `1073741894` — not producible by any of them — so the screenshot key is unpressable in a terminal.

That is a live gap rather than a hypothetical one:

- `TextRenderer.captureScreenshot()` is fully implemented (`src/rendering/textRenderer.py:356`) and writes a `.txt` capture, so the feature exists in text mode — it just has no reachable trigger.
- `WorldScreen.drawHelpOverlay()`s **text-mode** section explicitly advertises it: the HUD section renders `f"{keyName(screenshot)}  -  Take screenshot (.txt)"`, which displays as `PRINT SCREEN  -  Take screenshot (.txt)` to a terminal player who cannot press it.
- Three other screens also gate on the same binding — `inventoryScreen.py:101`, `statsScreen.py:43`, `chestScreen.py:274` — so all four screenshot call sites are dead in `--text`.

Per `CLAUDE.md`: "Every action that a terminal user needs must have an alt binding using an ASCII key. Add it to `DEFAULT_BINDINGS` and `ACTION_LABELS` in `src/config/keyBindings.py`, then check for it alongside the primary."

Suggested fix: add `"alt_screenshot": KeyCode.P` (P is unused across `DEFAULT_BINDINGS`) plus its `ACTION_LABELS` entry, check it alongside `screenshot` at all four call sites, and have the help overlay show the alt key name in text mode (the same `closeHelpAction` pattern `drawHelpOverlay` already uses for `toggle_help`). Update the README controls table too.

### LINKED ISSUE #538: Hotbar cycling ([ / ]) is hardcoded rather than a rebindable action, so it is missing from the Controls screen
In `WorldScreen._handleUtilityKey` (`src/screen/worldScreen.py:1163-1168`), hotbar cycling compares against raw `KeyCode` members instead of going through `KeyBindings`:

```python
elif key == KeyCode.LEFTBRACKET:
    current = self.player.getInventory().getSelectedInventorySlotIndex()
    self.changeSelectedInventorySlot((current - 1) % 10)
elif key == KeyCode.RIGHTBRACKET:
    current = self.player.getInventory().getSelectedInventorySlotIndex()
    self.changeSelectedInventorySlot((current + 1) % 10)
```

Every other action in that method resolves through `kb.getKey("<action>")`. Because these two do not, there are no `hotbar_cycle_left`/`hotbar_cycle_right` entries in `DEFAULT_BINDINGS` or `ACTION_LABELS` (`src/config/keyBindings.py`), which means:

- The Controls screen lists every action from `keyBindings.getActions()` — so hotbar cycling is invisible there and cannot be remapped, unlike all the other world-screen keys.
- The binding is not persisted or conflict-checked: `getConflictsForBindings` iterates the bindings dict, so a player who remaps another action onto `[` or `]` gets a silent double-fire with no red conflict warning.

The feature itself is real and documented — the README controls table lists ``` `[` / `]` | Cycle hotbar left / right ``` — and the **text-mode** branch of `WorldScreen.drawHelpOverlay()` lists `"[ ]  -  Cycle hotbar"`. The pygame branch of the same overlay does not mention it at all (it only shows `"1-0 / Scroll  -  Select hotbar slot"`), even though `[`/`]` work identically in pygame.

Suggested fix: add `"hotbar_cycle_left": KeyCode.LEFTBRACKET` / `"hotbar_cycle_right": KeyCode.RIGHTBRACKET` to `DEFAULT_BINDINGS` and `ACTION_LABELS`, switch the two branches to `kb.getKey(...)`, and render the key names in both help-overlay branches via the existing `keyName(action)` helper so pygame stops omitting the line.

dmccoystephenson:
## Summary

Three keybindings that the UI advertises but no player could actually trigger.

- **Controls screen `S: save` was dead code.** `ControlsScreen.handleKeyDownEvent` matched `S` in an earlier `(KeyCode.DOWN, KeyCode.S)` cursor-down branch, so the `elif key == KeyCode.S` save branch below it never ran — while the screen's own hint line reads `Up/Down: navigate - Enter/Space: remap - S: save - R: reset - Esc: cancel`. Saving otherwise needs the mouse "Save" button, and `TextInputSource.getMouseButtons()` returns `(False, False, False)` unconditionally, so a `--text` player could remap a binding but had **no way at all** to persist it (`Esc` runs `cancelAndReturn`). Dropped the `W`/`S` navigation aliases on this screen, leaving the arrow keys the hint already documents (both frontends decode them).
- **Screenshot was unreachable in `--text`.** `screenshot` was bound only to `KeyCode.PRINTSCREEN` (`1073741894`); `TextInputSource` can only emit `fromInt(ord(char))`, the four arrow escape sequences, and the Enter/Backspace control chars. `TextRenderer.captureScreenshot()` is fully implemented and the text-mode help overlay advertised the key. Added `alt_screenshot` (`P`) — including the missing `KeyCode.P = 112` member and its display name — and checked it alongside the primary at all four call sites (`worldScreen`, `inventoryScreen`, `statsScreen`, `chestScreen`).
- **Hotbar cycling was not a rebindable action.** `_handleUtilityKey` compared against raw `KeyCode.LEFTBRACKET`/`RIGHTBRACKET`, so `hotbar_cycle_left`/`_right` were in neither `DEFAULT_BINDINGS` nor `ACTION_LABELS`: invisible on the Controls screen, unremappable, and invisible to `getConflictsForBindings` (remapping another action onto `[` would silently double-fire with no red conflict warning). Registered both and routed the branches through `kb.getKey(...)`.

Both frontends handled: the help overlay now renders the cycle keys via `keyName()` in the pygame branch too (it had omitted the line entirely, even though `[`/`]` work there), and the text branch shows the `P` alt instead of `PRINT SCREEN`. README controls table and the text-mode tip updated.

+16 tests, including two drift guards: every `alt_*` binding must be terminal-typable (an arrow or ASCII `< 128`), and every action must have an `ACTION_LABELS` entry.

## Test plan

- [x] `python3 -m black --check src tests` — 260 files unchanged
- [x] `python3 -m pytest tests/` — **1094 passed** (was 1078)
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` — 247 passed
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` — 315 passed
- [x] New regression tests verified to fail against `main` (stash-checked): `test_hotbar_cycle_follows_a_remapped_binding` and `test_alt_screenshot_key_captures_a_screenshot`

Closes #536
Closes #537
Closes #538

---


dmccoystephenson:
## Self-review

Read the full diff against `main`. No blocking findings — the three fixes are independent, each has a regression test that was verified to fail on `main`, and both frontends are handled (the pygame help branch gained the hotbar-cycle line it had been missing, the text branch swapped Print Screen for the new ASCII alt). Notes below, in descending order of importance.

`src/rendering/keyCode.py:32` — Adding `P = 112` changes `fromInt(112)` from `None` to `KeyCode.P` process-wide, so the blast radius was traced rather than assumed inert. Every screen handler compares `key` against explicit members, and `None` matched nothing before, so the only new match is the `alt_screenshot` check itself. The one place worth checking was free-text entry: `SaveSelectionScreen.handleKeyDownEvent` short-circuits on `self.namingNewSave` before any shortcut branch (`saveSelectionScreen.py:237`), so typing a `p` in a save name still can't fire an action. `ControlsScreen`'s capture path also improves — `p` was previously rejected as "an unmodeled key can't be bound" and is now bindable like every other letter.

`src/screen/controlsScreen.py:241-247` — Dropping the `W`/`S` aliases is a small behavior removal, not just a reordering, so it deserves an explicit justification: the hint line this screen draws itself (`controlsScreen.py:~137`) documents `Up/Down` for navigation and never mentioned `W/S`, and the arrows work in both frontends (`TextInputSource._ARROW_SEQUENCES` decodes the CSI and SS3 forms). Reordering the branches instead — putting the `S` save branch first — would have been the alternative, but it would leave `S` doing two things depending on branch order, which is exactly the trap this fix exists to remove. `R` (reset) is unaffected: it never collided with a navigation branch.

`src/screen/worldScreen.py:1697` — The pygame help overlay's ACTIONS section goes from 7 to 8 items. The overlay is a fixed `y * 0.75` box with 22 px line spacing, and the content already runs close to the bottom edge on short windows (it overflowed at ~720 px display height before this change too, with 17 items). At the default window size — 90% of screen height, so ~970 px on a 1080 p display — the 18 items plus 4 headers still land well inside the box. Pre-existing and unchanged in kind, but flagging it since this PR moves one line closer to the edge; a maintainer smoke of F1 at a small window size would be the way to confirm, and if it does clip, the fix belongs in a separate scrolling/sizing change rather than here.

`tests/config/test_alt_keybindings.py` — The `test_every_alt_binding_is_typable_in_text_mode` guard encodes the rule as "an arrow member, or `int(key) < 128`". That is deliberately looser than what `TextInputSource` can literally produce (e.g. `KeyCode.SPACE`, `KeyCode.TAB` are under 128 but awkward as bindings) — the goal is only to catch the actual recurrence mode this PR fixes: an F-key, modifier, or Print Screen (all `> 1073741800`) being added as an `alt_` binding.

`src/screen/statsScreen.py:42-44` — black wraps this condition across the `getKey(` call, which reads worse than the sibling screens' one-liners. Left as-is rather than restructuring, since `statsScreen` is the only one of the four call sites without a local `kb = self.keyBindings` alias and introducing one purely for formatting would widen the diff.

One thing I checked and could not attribute to this PR: `python3 -m pytest tests/rendering/ tests/config/ tests/screen/ tests/ui/` in a single combined invocation fails 8 `tests/rendering/test_pygameInputSource.py` tests on pygame event-queue state. That reproduces identically on a clean `main` (10 failures there), passes when that file runs alone, and passes in the full `pytest tests/` run CI uses. Pre-existing test-isolation drift, not caused here, and not worth folding into this PR.

---

PR #546, opened 2026-08-02
test: make the integration harness's boot timeout overridable and its timeout failures self-explaining

### LINKED ISSUE #547: test: TextPlaythrough's 10s boot timeout has no environment override, so the integration suite can't run on a slower machine
### Problem

`TextPlaythrough.__init__` hardcodes `bootTimeout=10.0` (`tests/integration/textPlaythrough.py`). That value is the default for the *first* `expect` — the wait for the game subprocess to import pygame, boot and paint the main menu.

Two overrides exist, and neither is reachable from a `pytest` run:

- `runScript(path, ..., bootTimeout=...)` (`tests/integration/roamScript.py`) — but `test_roam_scripts.py` calls `runScript(scriptPath)` with no override.
- `roamScript.py --boot-timeout N` — standalone CLI only, one script at a time.

All four tests in `tests/integration/test_text_playthrough.py` construct `TextPlaythrough()` bare, so they are stuck at 10s too.

### Impact

On a host slower than the ubuntu CI runner the default is tuned for, every boot-dependent integration test fails identically:

```
E   AssertionError: timed out after 10.0s waiting for 'Roam'; last output:
```

Measured on an aarch64 Android/proot host: a quiescent boot-to-main-menu takes ~4.8-5.4s, but under the load of the rest of the suite the child's `import pygame` alone can exceed 30s, and 12 of the 15 tests in `tests/integration/` fail. There is no way to raise the budget short of editing the harness source.

### Proposal

Resolve the default in `TextPlaythrough.__init__` from a `ROAM_TEST_BOOT_TIMEOUT` environment variable when no explicit `bootTimeout` is passed, keeping 10.0 as the fallback so CI behaviour is unchanged. Both test modules go through that constructor, so it is a single point of control. Document it in `docs/roamscript.md`.

### LINKED ISSUE #548: test: a TextPlaythrough boot timeout reports an empty transcript and nothing about the game process
### Problem

When `TextPlaythrough.expect` times out it raises:

```python
raise AssertionError(
    "timed out after %.1fs waiting for %r; last output:\n%s"
    % (timeout, text, self.tail())
)
```

If the failure happened on the *first* `expect` — the boot wait — `self.tail()` is empty, because the child never painted anything. The message is then literally:

```
E   AssertionError: timed out after 10.0s waiting for 'Roam'; last output:
```

The harness knows more than that but never says it:

- `self._proc.poll()` would distinguish a child that exited from one still running.
- The child's stderr is captured, but only read in `__exit__` — after the assertion has already been raised and formatted. `cleanExit()`/`getStderr()` are consulted by `runScript` only once the script body has finished successfully, so on a failing `expect` they never run.

### Impact

A crashed child, a missing dependency, a child killed on startup and a child still importing pygame on a slow host all produce the identical, contentless message — and none of them is distinguishable from a genuine text-frontend rendering bug. Diagnosing one currently means editing the harness to print stderr.

### Proposal

Drain stderr non-blockingly (select on `self._proc.stderr.fileno()` with a zero timeout) and include the child's state plus any stderr it has already written in the `expect` timeout message, e.g. `child: still running, no stderr` or `child: exited with code 1, stderr: <traceback>`. `__exit__` must then append to `_stderr` rather than overwrite it, so text already drained for a failure message isn't lost.

dmccoystephenson:
## Summary

Two fixes to `tests/integration/textPlaythrough.py`, the driver behind both the `TextPlaythrough` tests and every `.roamscript`.

- **The 10s boot timeout was unreachable from a `pytest` run.** `TextPlaythrough.__init__` hardcoded `bootTimeout=10.0` — the wait for the game subprocess to import pygame, boot and paint its first screen. The two existing overrides don't help: `runScript(path, bootTimeout=...)` is never passed by `test_roam_scripts.py`, and `roamScript.py --boot-timeout` is a one-script standalone flag; all four tests in `test_text_playthrough.py` construct the harness bare. On a host slower than the ubuntu CI runner the default is tuned for, the suite simply couldn't be run — measured on an aarch64 Android/proot box, a quiescent boot takes ~5s but the child's `import pygame` alone can exceed 30s under the load of the rest of the suite, and 12 of the 15 tests in `tests/integration/` failed with `timed out after 10.0s waiting for 'Roam'`. New `resolveBootTimeout()` resolves explicit argument → `$ROAM_TEST_BOOT_TIMEOUT` → the unchanged `10.0` default, so CI behaviour is identical and one env var covers both modules. A malformed or non-positive env value falls back to the default rather than raising, so a shell typo can't masquerade as a game bug.
- **Those failures were undiagnosable.** `expect()` printed only `last output:` followed by nothing, because the child never painted. The child's stderr is collected in `__exit__`, long after the assertion is raised and formatted, and `cleanExit()`/`getStderr()` are only consulted after a successful body — so a crashed child, a missing dependency and one still importing pygame all looked identical. Added `describeChild()` plus a non-blocking `_drainStderr()`, folded into the timeout message: `child: still running, no stderr` / `child: exited with code 1, stderr: …`. `__exit__` now appends to `_stderr` instead of overwriting, so text drained for a failure message isn't lost.

Also simplified `runScript`'s `playthroughKwargs` dance (passing `bootTimeout=None` is now the natural default) and documented both changes in `docs/roamscript.md`, which previously mentioned the boot timeout only in passing and documented no way to change it under pytest.

Test-harness and documentation only — no game code changed, so the frontend-abstraction contract is untouched.

Closes #547
Closes #548

## Test plan

- [x] `python3 -m black --check src tests` — 270 files clean
- [x] `python3 -m pytest tests/` — **1147 passed** (+13 new)
- [x] `ROAM_TEST_BOOT_TIMEOUT=60 python3 -m pytest tests/integration/ -q` — 28 passed, vs. 12 failing on this host before the change
- [x] `python3 tests/integration/roamScript.py --boot-timeout 60 tests/integration/scripts/boot_to_main_menu.roamscript` — PASS (standalone CLI path unaffected)
- [x] New `tests/integration/test_text_playthrough_harness.py` covers the resolution order (explicit > env > default), the four malformed/non-positive env values, that the harness itself picks the env var up, and each `describeChild()` state including the stderr-quoting and `expect()`-message paths

---


dmccoystephenson:
## Self-review

Reviewed the full diff against `CLAUDE.md`'s conventions and this repo's harness contracts. No blocking findings; the notes below are the things I checked hardest and one clarification already pushed as `377df72`.

**Scope check — the frontend-abstraction rule doesn't apply here.** `CLAUDE.md`'s "handle both frontends" requirement covers drawing/input/HUD changes. This PR touches `tests/integration/textPlaythrough.py`, `tests/integration/roamScript.py`, a new test module and two docs; no `src/` file changed, no `Renderer`/`InputSource`/`Clock` call site moved, no `KeyCode` or `geometry.Rect` usage introduced. Confirmed with `git diff --stat main...HEAD`: `CHANGELOG.md`, `docs/roamscript.md`, and three files under `tests/integration/`.

**Findings, folded in as `path:line — note`:**

- `tests/integration/textPlaythrough.py:114` — `__exit__` changed from `self._stderr = ...` to `self._stderr += ...`. This is load-bearing and easy to miss: `describeChild()` may already have drained part of the stream, and an overwrite would silently discard exactly the text that explained the failure. The blocking `BufferedReader.read()` here is still correct after `_drainStderr`'s raw `os.read` on the same fd, because nothing ever reads through the buffered object before this point, so its buffer is empty and no bytes are stranded in it.
- `tests/integration/textPlaythrough.py:164` — `_drainStderr` reads `self._proc.stderr.fileno()` directly rather than through the `BufferedReader`, since the latter's `read()` blocks and the whole point is to report on a child that may still be running. `select(..., 0)` with a loop that returns on "not ready", `OSError` or empty chunk means it cannot block or spin. It is only ever reached from inside the context manager (via `expect`), so the fd is open.
- `tests/integration/textPlaythrough.py:187` — `describeChild()` is a new public method, so per `CLAUDE.md`'s testing rule it needs coverage: all four states are tested (not started / still running / exited with code / stderr quoted), plus the `expect()` message path.
- `tests/integration/textPlaythrough.py:36` — the env value is sanity-checked (malformed, `""`, `0`, negative all fall back) but an explicit argument is not. That asymmetry is deliberate — a machine-wide export is where a typo is likely, while `bootTimeout=0` from a caller could be an intentional fast-fail — but it was unstated, so `377df72` was pushed to say so in the docstring rather than leave a future reader guessing.
- `tests/integration/textPlaythrough.py:73` — the signature default moved from `10.0` to `None`. Nothing outside `runScript` constructs the harness with a positional second argument (checked: only `test_text_playthrough.py`'s four bare `TextPlaythrough()` calls and `roamScript.py`), so this is not a breaking change for any existing caller.
- `tests/integration/roamScript.py:200` — the `playthroughKwargs` conditional existed only to avoid passing `bootTimeout=None` into a constructor whose default was `10.0`. With `None` now the meaningful default, passing it straight through is equivalent and shorter. Verified the standalone CLI still works end to end: `python3 tests/integration/roamScript.py --boot-timeout 60 tests/integration/scripts/boot_to_main_menu.roamscript` → `PASS`.
- `docs/roamscript.md:32` — the sample failure message in the new section is the real output of the new format string (`"timed out after %.1fs waiting for %r; %s\nlast output:\n%s"`), not a paraphrase.

**CI behaviour is unchanged.** Neither `.github/workflows/tests.yml` nor `integration.yml` sets `ROAM_TEST_BOOT_TIMEOUT`, so CI keeps the same 10.0s budget it has today; the env var only helps a developer on slower hardware. `tests.yml` runs bare `python -m pytest`, so the new `test_text_playthrough_harness.py` is collected with no wiring.

**Doc accuracy pass** (per the dev-loop's Phase 7 table), done against the implementation rather than from memory: `README.md`'s controls table matches `DEFAULT_BINDINGS`/`ACTION_LABELS` including `alt_toggle_npc_mode`; `PLANNING.md`'s room types match `src/world/roomType.py` (6), its mob list matches `LIVING_ENTITY_TYPES` (6), and its recipe list matches `RecipeRegistry` (11, in the same order); `CLAUDE.md`'s frontend/keybinding contract still matches `src/rendering/` and `src/config/keyBindings.py`. No drift found, and this PR introduces none. `CHANGELOG.md` has the row for this change with the current passing count.

**Verification run on this host:** `black --check src tests` clean (270 files); `python3 -m pytest tests/` → 1147 passed under `ROAM_TEST_BOOT_TIMEOUT=90`. Worth stating plainly: on this host, *before* this change, 12 of the 15 tests in `tests/integration/` failed on the 10s boot wait, and 9 failed in a full-suite run — that is the environmental problem this PR is about, and it is what the env var fixes rather than papers over, since the default is untouched.

---

Issue #550, opened 2026-08-08
`--web` mode is broken: no COOP/COEP headers, and its WebSocket server has no client


dmccoystephenson:
## Summary

`python src/roam.py --web` starts `WebFrontend` (`src/rendering/webFrontend.py`), which brings up two listeners: an HTTP server on `webHttpPort` and a WebSocket server on `webWsPort`. Neither half of that pairing works any more, because `web/index.html` was converted into the Pyodide/Worker client and the server-side WebSocket frontend was left behind.

## Evidence

Gathered against `main` at 19dfece.

**1. The page served by `--web` contains no WebSocket client.**

`_startHttpServer` (`src/rendering/webFrontend.py:200-249`) serves `web/index.html` and rewrites the WebSocket port into it:

```python
html = html.replace(
    "const WS_PORT = 8765;",
    f"const WS_PORT = {wsPort};",
)
```

`web/index.html` no longer contains that string — nor `WebSocket`, nor any reference to `ws://`. The replacement is a silent no-op, and the WebSocket server started alongside it is never connected to. `WebSession`, `WebRenderer`, `WebInputSource` and the whole server-side render protocol are therefore unreachable through this entry point.

**2. The page that *is* served cannot run, because the required cross-origin isolation headers are missing.**

`web/index.html:428-437` allocates a `SharedArrayBuffer` for the input ring buffer, which browsers only permit on a cross-origin-isolated page. `web/serve.py` sends `Cross-Origin-Opener-Policy` / `Cross-Origin-Embedder-Policy` from its `end_headers` override; `_startHttpServer` has no such override.

Measured directly against both servers:

```
# _startHttpServer (the --web path)
/                     200 COOP= None        COEP= None
/web/game.zip         200 COOP= None        COEP= None
/web/game-worker.js   200 COOP= None        COEP= None

# web/serve.py (the Pyodide path)
/                     200 COOP= same-origin COEP= require-corp
/web/game.zip         200 COOP= same-origin COEP= require-corp
/web/game-worker.js   200 COOP= same-origin COEP= require-corp
```

**3. `--web` never produces the bundle the page downloads.**

`web/game.zip` is gitignored and is built by `web/build_zip.py`, which the `--web` path never invokes. On a fresh clone `/web/game.zip` returns 404 under `--web`; the `Dockerfile` runs `python3 web/build_zip.py` at build time precisely because of this, and then starts `python3 web/serve.py` — not `--web`.

[...]

PR #554, opened 2026-08-08
feat: add a `--help` usage summary and reject unknown arguments

### LINKED ISSUE #553: cli: `src/roam.py` has no `--help` and silently ignores unknown flags
## Summary

`src/roam.py` accepts three command-line flags — `--text`, `--web` and `--selftest` — but no usage output exists for any of them, and an argument that is not recognised is discarded without comment.

## Evidence

Checked against `main` at 171385d.

**1. No help flag is handled.**

Every read of the argument vector in `src/roam.py` is a membership test against one of three literals:

- `src/roam.py:13` — `if "--text" in sys.argv or "--web" in sys.argv:`
- `src/roam.py:255` — `if "--text" in argv:`
- `src/roam.py:275` — `if "--selftest" in argv:`
- `src/roam.py:280` — `if "--web" in argv:`

Neither `--help` nor `-h` appears anywhere in the file, and no `argparse`/`getopt` parser is constructed. `python src/roam.py --help` therefore falls through `main()` to `Roam(config, textMode=_shouldUseTextMode(argv))` and boots the game, which is the opposite of what the flag conventionally requests.

**2. An unrecognised argument is silently ignored.**

Because the flags are only ever tested for membership, anything else in `argv` is never examined. A mistyped `python src/roam.py --txt` on a headless machine boots into text mode by autodetection and looks like it worked; on a machine with a display it boots the graphical frontend, and the user is left to guess why the flag had no effect.

**3. The flags are not collected anywhere in the documentation either.**

`README.md` mentions `--text` twice in prose (a tip at line 40 and a run command at line 69) and `--selftest` once, in the Windows packaging section (line 114), as a way to verify a built bundle. No section lists the available options together, so there is no place — in the program or in the docs — where the full set can be read at once. (`--web` is undocumented on purpose; see #550.)

## Suggested resolution

- Handle `--help` / `-h` in `main()` by printing a usage block and returning `0`, before `Config()` is constructed so the flag works even when configuration cannot be loaded.
- Reject unknown arguments: print the offending argument and the usage block to stderr, and return a non-zero exit status.
- Keep the help path off the pygame display probe in `_earlyDetectTextMode`, so requesting usage neither initialises a display nor sets `LOG_FILE`.
- Add a short "Command-line options" subsection to `README.md`'s Run section listing the supported flags.

## Notes

`--web` should continue to be listed in the program's own usage text only if #550 is resolved in favour of repairing it; until then the usage text can describe the flags that work.

---

dmccoystephenson:
## Summary

- `src/roam.py` read its argument vector as three membership tests (`--text`, `--web`, `--selftest`) with no parser behind them. `--help` therefore fell through `main()` and booted the game, and any other argument was discarded without a word. The silent-ignore case is the more misleading of the two: on a headless host a typo such as `--txt` still lands in text mode by autodetection, so the run looks like the flag was honoured.
- `main()` now prints a usage summary and returns `0` for `-h`/`--help`, and prints each unrecognised argument plus the usage block to stderr and returns `2` otherwise. Both checks run ahead of `prepareWorkingDirectory()`/`Config()`, so usage stays reachable when configuration cannot be loaded.
- The module-level display probe is skipped on the help path, so asking for usage neither initialises a pygame display nor sets `LOG_FILE=roam.log`.
- The usage line and the error prefix are taken from `os.path.basename(argv[0])` rather than a hardcoded script name, because the same `main()` runs inside the packaged builds — `Roam.exe --selftest` is already documented, so `Roam.exe --help` printing "Usage: roam.py" would name a file that build does not contain.
- Two arguments are deliberately not rejected: `-psn_<n>_<n>`, which Finder can hand to a launched macOS `.app` bundle (this repo ships one via `macos-package.yml`), and `--web`, which stays accepted but is left out of the usage text while #550 has it broken. A test asserts the usage text omits `--web`, so repairing that mode will surface this as a failing test rather than as forgotten documentation.
- `README.md` gains a "Command-line options" table under Run; `CHANGELOG.md` gains a summary row and a session entry.

## Frontend abstraction

No drawing, input or HUD code is touched — the change is confined to argument handling that runs before any frontend is constructed, so the pygame and text paths are affected identically. `--text` remains a recognised flag and `tests/integration`'s harness, which launches `roam.py --text`, is unaffected.

## Test plan

- [x] `python3 -m pytest -q --ignore=tests/integration` → 1147 passed
- [x] `tests/integration` → 28 passed, run under `ROAM_TEST_BOOT_TIMEOUT=120`. This host is slower than the 10s default `resolveBootTimeout()` assumes and produces spurious boot timeouts at the default; that is the pre-existing limitation the variable was added for on 2026-08-02, not a change from this diff. Total 1175, up from 1165 on `main` (+10)
- [x] `python3 -m black --check src tests` → 272 files unchanged
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` → 247 passed (text-interface unit tests)
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` → 315 passed (pygame-facing screen tests)
- [x] `python3 src/roam.py --help` → usage on stdout, exit `0`, and `git status --porcelain` clean afterwards (no `roam.log` created)
- [x] `python3 src/roam.py --txt` → `roam.py: unrecognized argument: --txt` plus usage on stderr, exit `2`

Closes #553

---


dmccoystephenson:
## Self-review

No configured reviewer exists on this repository, so the diff was reviewed inline against `CLAUDE.md` and the repo's own conventions. One defect was found and has been fixed on the branch (5c99366); the rest of the notes below are confirmations rather than requested changes. Findings are given as `path:line — finding` because this session cannot create anchored review comments.

### Fixed during review

`CHANGELOG.md:11` and `CHANGELOG.md:164` — both referred to a helper named `_USAGE` and omitted `_programName()`/`_usage()`. Those were introduced later on this branch, when the usage text was made to follow `argv[0]`; the entry therefore described an intermediate state of its own diff. Corrected in 5c99366.

### Confirmed correct

`src/roam.py:73-76` — the `if _wantsHelp(sys.argv): pass / elif _earlyDetectTextMode() ...` guard reads oddly at first, but the empty branch is what makes the help path skip the pygame display probe entirely, which is the point of the change. Rewriting it as a single `not _wantsHelp(...) and ...` condition would hide that intent behind an operator. Verified by hand: `python3 src/roam.py --help` in a clean tree leaves `git status --porcelain` empty, so no `roam.log` is created.

`src/roam.py:24-27` — `os.path.basename` is used on `argv[0]`, which is correct on both platforms that ship a packaged build (`ntpath` splits on `\` when running on Windows). The corresponding test uses a forward-slash path, so it does not depend on the separator of the host running the suite.

`src/roam.py:38-49` — `-psn_<n>_<n>` is skipped rather than rejected. This is not speculative: `macos-package.yml` builds a `.app` bundle, and Finder can pass that argument to a bundle it launches. Without the exemption, a Gatekeeper-launched `Roam.app` could exit `2` before reaching the game.

`src/roam.py:12` — `--web` is absent from the usage text while #550 has that mode broken end to end. `tests/test_roam.py:114` locks this in, so repairing #550 will fail that test rather than silently leaving the flag undocumented.

`src/roam.py:321-332` — both checks precede `prepareWorkingDirectory()` and `Config()`, so usage remains reachable when `config.yml` is missing or malformed. Exit status `2` for a usage error matches the convention `argparse` establishes.

### Frontend abstraction

No `Renderer` / `InputSource` / `Clock` call, no drawing, no `KeyCode` comparison and no `getGameAreaRect()` use appears in the diff. The changed code runs before any frontend is constructed, so the pygame and text paths are affected identically and the "handle both frontends" rule has nothing to bind to here. `--text` remains a recognised flag, which is what `tests/integration/textPlaythrough.py:94` launches.

### Test coverage

Every new public-facing behaviour has an anchor: both help spellings, the usage text's omission of `--web`, single and multiple unknown arguments, stream routing (stdout for help, stderr for errors), exit statuses, help winning over an unknown argument, `argv[0]` never being read as a flag, the macOS process-serial-number exemption, and the program name following `argv[0]`.

### Documentation

`README.md`'s new "Command-line options" table was checked against `_USAGE_TEMPLATE` and against `runSelfTest()` — the `--selftest` description matches what that function actually does (`pygame.init()`, load `assets/images/player_down.png`, parse `schemas/tick.json`, construct `Config()`). The `#building-a-standalone-executable-advanced` anchor was checked against the heading list in the file.

---

PR #558, opened 2026-08-09
feat: Per-level map images so cave levels no longer overwrite the surface map

### LINKED ISSUE #557: feat: per-level maps — cave levels overwrite the surface map and the minimap ignores depth
## Summary

The map is currently flat: one stitched image is kept for the whole world, with no notion of depth. Since caves were added, surface rooms and cave rooms at the same `(x, y)` write to the same map tile and overwrite one another. It is proposed that each level of the world carry its own map, and that the map shown and updated always be the one for the level the player currently occupies.

## Current behaviour

Observed against `main` at 9c27ed5.

Rooms are keyed by `(x, y, z)` throughout the world model (`src/world/map.py:18-119`), and room save files already carry depth in their filename (`src/config/config.py:340-347`). The map pipeline does not:

**1. Room PNGs are named without `z`.**

`saveCurrentRoomAsPNG` (`src/screen/worldScreen.py:1300-1335`) writes to `<saveDir>/roompngs/{x}_{y}.png`, and `isCurrentRoomSavedAsPNG` (`:1289-1298`) checks the same path. A cave room at `(0, 0, -1)` therefore writes over the surface room at `(0, 0)`.

**2. The pending-write guard is also flat.**

`roomKey = (self.currentRoom.getX(), self.currentRoom.getY())` (`src/screen/worldScreen.py:1301`) — two rooms at the same coordinates on different levels share one key, so a queued write for one can suppress the write for the other.

**3. The stitcher parses only two coordinates.**

`MapImageGenerator.generateMapImage` splits the filename into exactly `x` and `y` (`src/mapimage/mapImageGenerator.py:87-89`) and pastes every image into a single `mapImage.png`. Adding a `_z` suffix to the filename would break this parser, so the change has to be made on both sides.

**4. The minimap loads one image and one image only.**

`drawMiniMap` (`src/screen/worldScreen.py:1403-1470`) caches and draws `mapImage.png` regardless of `self.currentZ`, even though the depth indicator directly beside it reports the current level (`:1856-1880`).

**5. The text minimap flattens all levels.**

`_buildTextMinimapRows` (`src/screen/worldScreen.py:1354-1374`) builds its known-room set as `{(room.getX(), room.getY()) for room in self.map.getRooms()}`, with no `z` filter. Visiting a cave room at `(1, 1, -1)` marks the surface room at `(1, 1)` as explored.

The visible result is a map that mixes stone and grass at random depending on which level happened to be rendered most recently, and a minimap that does not correspond to the level the player is standing on.

## Proposed behaviour

- Each `z` level should have its own map image, generated and stored independently.
- The minimap — graphical and text — should display the map for the player's current level, and switch as the player descends and ascends.
- Room capture should be per level, so that visiting a cave never disturbs the surface map and vice versa.
- The depth indicator and the map should always agree about which level is being shown.

## Implementation notes

- **Filenames** — either a `{x}_{y}_{z}.png` scheme with a parser that treats a missing third component as `z = 0` (matching how `getRoomFilePath` handles legacy surface saves at `src/config/config.py:340-347`), or a per-level subdirectory such as `roompngs/z0/`, `roompngs/z-1/`. The subdirectory form keeps `MapImageGenerator`'s existing two-component parser intact and gives it a natural per-level invocation.
- **Outputs** — one stitched image per level, e.g. `mapImage_z0.png` / `mapImage_z-1.png`. `MapImageUpdater` (`src/mapimage/mapImageUpdater.py`) currently regenerates a single image on a 300-tick cooldown and calls `clearRoomImages()` afterwards; both need to become level-aware so that clearing one level's sources does not discard another's.
- **Room key** — `_pngSavePending` should key on `(x, y, z)`.
- **Minimap cache** — the cached surface in `drawMiniMap` must be invalidated on a `z` change, otherwise the previous level's image persists after descending.
- **Text minimap** — the known-room set needs filtering to `self.currentZ`.
- **Existing saves** — saves made before the change contain flat `{x}_{y}.png` files that may hold cave content stitched over surface content. Treating unsuffixed files as `z = 0` is the simplest migration; whether the stale mixed tiles are worth clearing on load is a decision to make.
- **Frontends** — per `CLAUDE.md`, both the pygame and text paths must be handled. PNG capture is a no-op under the web renderer (`src/rendering/webRenderer.py:106`).
- **Tests** — `tests/screen/test_worldScreen_minimap.py` covers the current flat behaviour and will need extending; coordinate and bounds handling in `MapImageGenerator` has existing coverage to build on.

## Open questions

- Should a key be added to page through the maps of other levels, as the cave design document suggests ("separate minimap layer per depth, toggled with a key" — `docs/cave-generation-design.html:576`), or should the map always be locked to the player's current level?
- Should levels the player has never visited be shown at all, or hidden entirely?

Related: player-built ladders between levels are filed separately as #556.

---

dmccoystephenson:
## Summary

The map was flat while the world has been keyed by `(x, y, z)` since caves landed. Room captures were written to `roompngs/{x}_{y}.png` and `MapImageGenerator` split that filename into exactly two coordinates, so the cave room at `(0, 0, -1)` overwrote the surface room's capture at `(0, 0)` and both were stitched into a single `mapImage.png`. The minimap therefore showed stone or grass depending on which level had been rendered most recently, while the depth indicator beside it correctly reported the level.

Three further flat spots compounded it and are fixed alongside:

- `_pngSavePending` keyed on `(x, y)`, so a queued write for one level suppressed the other level's write at the same coordinates.
- `clearRoomImages()` deleted every level's captures after stitching, discarding captures another level was still waiting on.
- `_buildTextMinimapRows` built its known-room set with no `z` filter, so visiting the cave room at `(1, 1, -1)` marked the surface room at `(1, 1)` as explored.

## What changed

- **`src/mapimage/mapImagePaths.py` (new)** — `getMapImageFilename`, `getRoomImageFilename` and `parseRoomImageFilename` in one place. The convention follows `Config.getRoomFilePath`: `z = 0` stays unsuffixed and everything else is suffixed (`0_0.png` / `0_0_-1.png`, `mapImage.png` / `mapImage_z-1.png`). Saves made before this change are therefore read as surface data with no migration step.
- **`MapImageGenerator`** — `generate`, `getRoomImages`, `clearRoomImages`, `getMapImagePath` and `mapImageExists` all take a level; `getRoomImagePath` and `getLevelsWithRoomImages` were added; the coordinate parser is shared with the new module and skips filenames it does not recognise.
- **`MapImageUpdater`** — every level with captures waiting is stitched into its own image, so descending mid-cooldown no longer strands the surface's pending captures.
- **`WorldScreen`** — the capture path and the pending-write key both carry the level (`getCurrentRoomImagePath` was extracted so the path is built once); `drawMiniMap` reads the current level's image and drops its cached surface on a `z` change, without which the previous level's frame persisted for up to 60 ticks after descending; the text minimap's known-room set is filtered to `self.currentZ`.

## Frontends

Both paths are covered. The graphical minimap follows the current level and invalidates on change; the text minimap's room grid is level-scoped. `_drawDepthIndicator` already named the level in both frontends, so map and indicator now agree in either. No new keybinding is introduced, so no `alt_` binding is needed.

## Open questions from the issue

The map is locked to the player's current level rather than gaining a paging key; the paging key is left as the remaining part of the design document's z-aware-minimap item, and `docs/cave-generation-design.html` was updated to record what is now done and what is not. Levels never visited have no map image and are simply not drawn.

## Test plan

- [x] `python3 -m black --check src tests` — 275 files unchanged
- [x] `python3 -m pytest tests/` — 1202 passing (+27)
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` — 247 passed
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` — 324 passed
- [x] `python3 -m pytest tests/integration -q` — 27 of 28 passed under a raised `ROAM_TEST_BOOT_TIMEOUT`; the remaining one is a boot-speed timeout on this host that reproduces identically on unmodified `main` and passes in isolation on both

⚠️ A live pygame smoke is recommended before merge: descend into a cave and confirm the minimap switches to the cave map and back on ascending.

Closes #557


dmccoystephenson:
## Self-review

The full diff was read against this repo's conventions (frontend abstraction, keybindings, tests, doc drift). One finding was acted on before this comment was posted; the rest are recorded as accepted trade-offs or verified non-issues.

### Acted on

**`src/screen/worldScreen.py:1317` / `src/mapimage/mapImageUpdater.py:55` — the capture path was assembled in two places.** As first pushed, `MapImageGenerator.getRoomImagePath` was added but never called by production code, while `WorldScreen` built the same `roompngs/{name}` string by hand and repeated the `"/roompngs"` literal for its `os.makedirs`. That is exactly the drift the new `mapImagePaths` module exists to prevent, one layer up. Fixed in ad250b7: `MapImageUpdater` gained `getRoomImagePath`, `getRoomImagesDirectoryPath` and `getMapImagePath` as thin delegates, `WorldScreen` now asks for all three, and the unused public method is no longer unused. The delegate placement follows the `roompngsLock` these two classes already share rather than reaching two levels deep through `mapImageUpdater.mapImageGenerator`.

### Accepted trade-offs

**`src/mapimage/mapImageGenerator.py:63` — `clearRoomImages(z)` leaves unrecognised filenames behind.** The no-argument form still deletes everything in `roompngs/`, matching the previous behaviour, but the level-scoped form skips any file `parseRoomImageFilename` rejects. A stray file therefore accumulates rather than being swept, since `getLevelsWithRoomImages` will not report it either. This is deliberate: deleting a file that matches no known pattern is not something a call scoped to one level should be doing, and only the game writes into that directory.

**`src/mapimage/mapImageUpdater.py:66` — no map image is written when no level has captures waiting.** Previously `generate()` re-saved the canvas unconditionally on every update. The loop now does nothing in that case. The re-save was idempotent, `drawMiniMap` already handles a missing file by returning early, and the exit path in `WorldScreen.saveAndWait` captures the current room and flushes the executor before calling the updater, so a real session always has at least one level to stitch.

### Verified, not defects

- **`_miniMapCachedZ` starts at 0 while a save can be resumed underground** (`src/screen/worldScreen.py:128`). The first `drawMiniMap` after such a load sees a mismatch and takes the invalidation branch, which is harmless against an already-empty cache and leaves the field correct from then on.
- **The 60-tick reload window does not delay a level change** (`src/screen/worldScreen.py:1441`). Invalidation clears `_cachedMiniMapImage`, and the reload condition is satisfied by `is None` before the tick comparison is reached.
- **`room.getZ()` is safe on every element of `map.getRooms()`** (`src/screen/worldScreen.py:1385`). `Room` defines it at `src/world/room.py:52` and `Map` keys its index on it, so no room can reach the minimap without one.
- **The parser rejects what it should.** `mapImage.png`, `a_b.png`, `0_0_x.png` and `0_0_-1_extra.png` all return `None`; negative coordinates round-trip in both the two- and three-component forms. Covered by `tests/mapimage/test_mapImagePaths.py`.

### Frontends and docs

Both rendering paths are handled, as `CLAUDE.md` requires: the graphical minimap follows the current level and drops its cache on a change, and the text minimap's known-room set is filtered to the current level. `_drawDepthIndicator` already branched on `supportsImageLoading()` and named the level in both frontends, so map and indicator agree in either. No new key is introduced, so no `alt_` binding is required. `README.md`'s controls table and `PLANNING.md`'s room-type, mob and recipe lists are unaffected; `docs/cave-generation-design.html` listed a z-aware minimap as a future extension and was updated to record the per-depth layers as done and the paging key as the remainder.

### Tests

1202 passing (+27), `black --check` clean. The graphical-minimap change is still the kind that unit tests cannot fully vouch for, which is why the CHANGELOG row carries the ⚠️ visual flag and this PR asks for a descend-and-ascend smoke before merge.

This review was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

dmccoystephenson:
One open question from #557 was left unanswered by the description above: what becomes of saves made before this change, whose flat `{x}_{y}.png` captures and `mapImage.png` may already hold cave content stitched over surface content.

No migration step is performed. Unsuffixed files are read as surface data, so such a save opens with its existing map intact, mixed tiles and all. Those tiles then heal on their own as the world is played: `clearRoomImages(z)` empties that level's captures after each stitch, which makes `isCurrentRoomSavedAsPNG()` false again, so `draw()` recaptures whichever room the player is standing in and the fresh capture is pasted over the stale tile on the next update. Rooms that are never revisited keep whatever was baked in.

Clearing the stale tiles eagerly on load was considered and not done: it would discard correct surface data along with the mixed data, since nothing on disk distinguishes the two, and the mixing only ever affected coordinates visited at more than one level.

PR #563, opened 2026-08-13
feat: page the minimap through levels the player is not standing on

### LINKED ISSUE #559: feat: a key to page through the maps of levels the player is not standing on
## Summary

With per-level map images in place (#557, PR #558), each level of the world now stitches its own map image and the minimap follows the level the player occupies. The remaining half of the cave design document's z-aware-minimap item is a way to look at a level the player is not standing on.

## Current behaviour

Observed against `feat/per-level-map-images` at ad250b7.

- Captures are written per level as `roompngs/{x}_{y}.png` for the surface and `roompngs/{x}_{y}_{z}.png` below it (`src/mapimage/mapImagePaths.py`).
- `MapImageUpdater._doUpdateMapImage` stitches every level that has captures waiting, each into `mapImage.png` / `mapImage_z{z}.png` (`src/mapimage/mapImageUpdater.py:66-79`).
- `WorldScreen.drawMiniMap` reads `self.mapImageUpdater.getMapImagePath(self.currentZ)` and drops its cached surface whenever `self.currentZ` changes, so the drawn map is always the current level's and nothing else is reachable.
- The text minimap filters its known-room set to `self.currentZ` in `_buildTextMinimapRows`, with the same consequence.

The maps of other levels therefore exist on disk and are kept up to date, but no input path reaches them. A player standing on the surface cannot check the layout of the cave below without descending, and a player at depth 3 cannot see how the shallower levels connect.

## Proposed behaviour

A key that shifts the level being *displayed* without moving the player, plus an indication that the map on screen is not the level being stood on. Returning to the player's own level should be possible in one press, and the deepest generated level is `z = -3` (`src/screen/worldScreen.py:959-961`), so the range is small and bounded.

## Implementation notes

- **Displayed level** — a field distinct from `self.currentZ`, reset to it on `_descend()` / `_ascend()` so paging never survives an actual move between levels. `drawMiniMap` and `_buildTextMinimapRows` would read the displayed level rather than `self.currentZ`.
- **Cache** — `drawMiniMap` already invalidates `_cachedMiniMapImage` on a level change by comparing against `_miniMapCachedZ` (`src/screen/worldScreen.py:1441-1447`); that comparison would move to the displayed level and needs no other change.
- **Levels with no map** — a level never visited has no `mapImage_z{z}.png`, and `drawMiniMap` returns early on a missing file. Whether paging should skip such levels or show an explicit "unexplored" state is a decision to make.
- **Keybindings** — per `CLAUDE.md`, any new key needs an entry in both `DEFAULT_BINDINGS` and `ACTION_LABELS` in `src/config/keyBindings.py` and an ASCII `alt_` binding so a terminal player can reach it, checked alongside the primary in `_handleUtilityKey`. Note that `minimap_zoom_in` / `minimap_zoom_out` / `toggle_minimap` already occupy `=`, `-` and `M`.
- **Frontends** — both paths need it: the graphical minimap and the text minimap's room grid, plus whatever marks the displayed level as not the player's own. `_drawDepthIndicator` already branches on `supportsImageLoading()` and reports the level in both frontends, so it is the natural place for that marking.
- **Tests** — `tests/screen/test_worldScreen_minimap.py` covers the per-level selection and the cache invalidation and is where the paging behaviour belongs alongside them.

## Open questions

- Should paging be limited to levels the player has actually visited, or should any level within the `z = -3` floor be reachable?
- Should the displayed level reset when the minimap is toggled off and on again, or persist for the session?

Related: the per-level map images this builds on are #557; player-built ladders between levels are #556.

### LINKED ISSUE #562: Controls screen shows "KeyCode.F2" instead of "f2" for Toggle NPC/CPC Mode
## Summary

The Controls screen renders the raw enum repr `KeyCode.F2` in the key column of the **Toggle NPC/CPC Mode** row, rather than the human-readable `f2` shown for every other binding.

## Cause

`displayName()` in `src/rendering/keyCode.py` falls back to `str(keyCode)` when a member is absent from the `_DISPLAY_NAMES` table:

```python
def displayName(keyCode):
    if keyCode is None:
        return "None"
    return _DISPLAY_NAMES.get(keyCode, str(keyCode))
```

`KeyCode.F2` was added to `KeyBindings.DEFAULT_BINDINGS` (as `toggle_npc_mode`) without a matching `_DISPLAY_NAMES` entry — `KeyCode.F1` and `KeyCode.F3` both have one. It is the only bound key in the table that is missing, which was confirmed by walking `DEFAULT_BINDINGS` against `_DISPLAY_NAMES`.

`str()` on an `IntEnum` returns `'KeyCode.F2'` under Python 3.10 (the interpreter this repository is developed against) and `'1073741883'` under Python 3.11+, so the rendered text is wrong either way.

## Where it shows

`src/screen/controlsScreen.py:99` computes `keyName = displayName(...)` for every visible action and draws it in the key button, so the Controls screen is the visible surface. `KeyBindings.getKeyName()` delegates to the same function.

## Suggested resolution

Add `KeyCode.F2: "f2"` to `_DISPLAY_NAMES`, and cover the invariant with a test asserting that every key in `DEFAULT_BINDINGS` has a `_DISPLAY_NAMES` entry, so a future binding cannot reintroduce the gap.

---

dmccoystephenson:
## Summary

- A displayed minimap level, `_miniMapViewZ`, was added to `WorldScreen`, distinct from the player's own `currentZ` and resolved through the new `getMiniMapDisplayZ()`. `None` means "follow the player", so a save loaded underground needs no synchronisation, and `_descend()` / `_ascend()` reset to `None` so paging never survives a real move between levels.
- Three actions were added for paging it: **Page Up** (level above), **Page Down** (level below) and **Home** (back to the player's own level, in one press), each with the ASCII alt binding a terminal player needs — `,`, `.` and `/` respectively. `KeyCode` gained the six corresponding members with display names.
- Both frontends were updated. The graphical minimap loads the displayed level's stitched image (the existing `_miniMapCachedZ` invalidation compare moved to it unchanged) and gains a caption naming the level and the key that returns; the text minimap shows the displayed level's known rooms, marks the center cell `+` rather than the facing arrow — the player is not on that level, so the marker means "directly above or below you" — and gains the same caption as a second header line.
- A level with no stitched map is drawn as an explicit "unexplored" panel rather than nothing. On the player's own level a blank minimap stays as it was, since it fills in on its own as rooms are captured; on a level paged to deliberately, drawing nothing would be indistinguishable from a key that did nothing.
- The `-3` floor `_descend()` had inline was extracted to `DEEPEST_Z` and is now shared with the paging clamp.
- `KeyCode.F2` had no `_DISPLAY_NAMES` entry, so `displayName()` fell through to `str()` and the Controls screen rendered `KeyCode.F2` in the Toggle NPC/CPC Mode row. The entry was added and the invariant is now covered by a test walking `DEFAULT_BINDINGS` against the table. It was found while the paging key codes were being added.

## Decisions on the issue's open questions

- **Which levels are reachable** — the full generated range (`0` down to `DEEPEST_Z`), not only visited levels. The range is small and bounded, and the unexplored panel answers "what is down there" more usefully than a keypress that silently skips.
- **Whether the displayed level persists** — it is discarded when the minimap is hidden, so re-enabling it always starts on the player's own level. It also resets on any descent or ascent, as the issue specifies.
- Paging while no minimap is on screen reports "Minimap is off" and changes nothing, so no invisible state can accumulate. Visibility is judged by `_isMiniMapVisible()`, which mirrors both branches of the draw call — the text minimap is drawn even while the toggle is off, so a `--text` player can still page the grid in front of them.

## Test plan

- [x] `python3 -m black --check src tests` — 276 files unchanged
- [x] `python3 -m pytest tests/` — 1229 passed (+24)
- [x] `python3 -m pytest tests/rendering/ tests/config/ -q` — text-interface unit tests pass
- [x] `python3 -m pytest tests/screen/ tests/ui/ -q` — pygame-facing screen tests pass
- [x] `python3 tests/integration/roamScript.py tests/integration/scripts/minimap_level_paging.roamscript` — PASS; the new script drives `,` `.` `/` end to end in the real text frontend, covering the caption, both clamps and the one-press return
- [ ] ⚠️ Live pygame smoke recommended before merge: page below the surface and confirm the caption and the unexplored panel render where expected

## Documentation

`README.md`'s controls table and its terminal-mode tip, the in-game help overlay (both frontends), `CHANGELOG.md` and the z-aware-minimap entry in `docs/cave-generation-design.html` were all updated.

Closes #559
Closes #562


dmccoystephenson:
## Self-review

The full diff was read against this repository's conventions (`CLAUDE.md`'s frontend-abstraction and keybinding rules, the Phase 7 documentation table). Two defects were found and have been fixed in `5b61cd3` on this branch; the remaining notes are observations that were deliberately left alone.

### Findings fixed on the branch

`src/screen/worldScreen.py:1467` — the paging keys were gated on `self.config.showMiniMap`, but the text minimap is drawn regardless of that toggle: the draw site reads `if self.config.showMiniMap and self.minimapScaleFactor > 0: self.drawMiniMap()` / `elif not self.renderer.supportsImageLoading(): self._drawTextMinimap()`. A `--text` player who had toggled the minimap "off" would therefore still see the room grid while the paging keys answered "Minimap is off" and did nothing. `_isMiniMapVisible()` was added, mirroring both branches of that draw call, and both `_pageMiniMapLevel` and `_resetMiniMapLevel` now consult it. Covered by `test_paging_works_in_text_mode_even_with_the_minimap_toggled_off`.

`src/screen/worldScreen.py:1428` — the status message named the primary reset binding unconditionally, so a terminal player was told "HOME to return" with a key a terminal cannot send. `_getMiniMapResetKeyName()` now picks `alt_minimap_level_reset` in text mode, which is what the on-screen caption was already doing through `_buildMiniMapViewLabel`. The duplicated selection between the two call sites was folded into that one helper. Covered by `test_paging_status_names_the_ascii_return_key_in_text_mode`.

### Reviewed and considered correct

`src/screen/worldScreen.py:1573` — the cache-invalidation compare was repointed from `self.currentZ` to the displayed level, so a stale frame from another level cannot be drawn: `_cachedMiniMapImage` is cleared on every change of displayed level, which means the "file missing but a cached image exists" branch is unreachable immediately after paging and the unexplored panel is what actually renders.

`src/screen/worldScreen.py:1587` — the unexplored panel is drawn only when the displayed level is not the player's own. On their own level a missing map stays silent, which preserves the existing behaviour of a minimap that fills in as rooms are captured.

`src/screen/worldScreen.py:1002` and `:1029` — `_miniMapViewZ` is reset before `_loadOrGenerateRoom`, so no draw can observe a half-changed level. Storing `None` rather than the numeric level for "the player's own" is what makes a save loaded underground need no synchronisation at `initialize()` time.

`src/config/keyBindings.py:46` — `,`, `.` and `/` are ASCII and reach `TextInputSource` through `fromInt(ord(char))`, and `test_alt_bindings_have_no_conflict_with_primary` confirms none of the six new keys collide with an existing binding.

One residual edge case was left as-is: a paged-to level whose map file exists but fails to load (corrupt PNG) still returns without drawing anything rather than showing the unexplored panel, because "unexplored" would be the wrong explanation for a corrupt file. That path already logs a warning once on the good-to-failed transition.

### Documentation

`README.md` (controls table and the terminal-mode tip), the in-game help overlay in both frontends, `CHANGELOG.md`, and the z-aware-minimap entry in `docs/cave-generation-design.html` were each checked against the implementation after the fixes above. No further drift was found.

Test counts in the PR body and the changelog row were updated for the two added tests: 1229 passing.

Issue #556, opened 2026-08-08
feat: craftable ladders so players can build their own way between cave levels


dmccoystephenson:
## Summary

Vertical movement through the world is currently possible only at the points the world generator chose. It is proposed that ladders become a craftable, placeable item, so that a player can build their own way down into the caves and back up again, rather than being restricted to generated links.

## Current behaviour

Observed against `main` at 9c27ed5.

The z-axis machinery already exists — rooms are keyed by `(x, y, z)` in `Map` (`src/world/map.py:18-119`), caves are generated for `z < 0` (`src/world/roomFactory.py:164-238`), and `WorldScreen` tracks `self.currentZ` with `_descend()` / `_ascend()` (`src/screen/worldScreen.py:958-1000`). What is missing is any player agency over where the links between levels are.

- **Descending** is only possible on a `CaveEntrance`. Those are placed by the generator: on the surface with a per-biome chance (mountain 40%, forest 15%, grassland 5% — `src/world/roomFactory.py:102-139`, `:330-360`), and inside a cave room with a 30% / 20% chance of a deeper entrance at depths 1 and 2 (`src/world/roomFactory.py:172-238`).
- **Ascending** is only possible on the single `CaveLadder` that `_generateCaveTiles` places once per cave room at a random open cell far from the entry point.
- Neither entity is in `PICKUPABLE_TYPES` (`src/screen/pickupableEntities.py`), so neither can be gathered and relocated.
- No recipe produces either one; `RecipeRegistry` holds nine hardcoded recipes and none of them yield a ladder (`src/crafting/recipeRegistry.py:23-88`).

The consequence is that a cave level can be entered and left only through generated geometry. If a player digs out a passage or wants a shortcut back to the surface from a deep room, no means of building one is available.

## Proposed behaviour

A ladder should be craftable from wood, placeable like any other item, and gatherable again once placed. Placing a ladder should establish a usable vertical link at that tile:

- Placing a ladder while underground should provide a way **up** to `z + 1`, matching how a generated `CaveLadder` behaves today.
- Placing a ladder on the surface, or in a cave with no generated deeper entrance, should provide a way **down** to `z - 1`, so that descent is no longer gated on the generator's dice rolls. The existing depth floor of `z = -3` (`src/screen/worldScreen.py:959-961`) should continue to apply.

## Implementation notes

[...]

Issue #551, opened 2026-08-08
docs: agent instruction files have drifted from the source tree


dmccoystephenson:
## Summary

Both agent-loaded instruction files — `.github/copilot-instructions.md` and `CLAUDE.md` — carry claims that no longer match the source tree. Because these files are read at the start of every agent session, the drift is inherited by every subsequent change.

They are filed together in one issue rather than fixed as part of a documentation sweep because edits to agent-loaded configuration require explicit human authorization, and `.github/copilot-instructions.md` is on the do-not-auto-merge path list.

## Drift in `.github/copilot-instructions.md`

Each item was checked against `main` at 19dfece.

| Claim | Reality |
|---|---|
| "**Version:** 0.11.0-SNAPSHOT (tracked in `version.txt`)" | `version.txt` contains `0.12.0-SNAPSHOT` |
| "**Game framework:** Pygame 2.1.2" | `requirements.txt` specifies `pygame>=2.6.1`; the exact-version pin was deliberately removed (see the comment at the top of `requirements.txt`) |
| Technology stack omits structlog and websockets | `requirements.txt` lists `structlog>=25.0.0` and `websockets>=12.0`, and `LOGGING.md` documents structlog as the logging backend |
| "interact with living entities (chickens, bears)" | `src/entity/living/livingEntityRegistry.py` registers Bear, Chicken, Deer, Rabbit, Snake and Wolf |
| Repository layout lists the `src/` subdirectories | `src/api/`, `src/goals/`, `src/npc/`, `src/rendering/` and `src/update/` are all absent from the list, as are `src/appPaths.py`, `src/jsonPersistence.py`, `src/schemaCache.py` and `src/textDemo.py`. Top-level `web/`, `docs/`, `deploy/` and `scripts/` are also unlisted |
| "`pytest.ini` adds `.`, `src`, and `src/entity` to `pythonpath`" | `pytest.ini` also adds `tests/integration` and `web` |
| CI "Installs dependencies (`pygame`, `pytest`, `pytest-cov`, `jsonschema`, `Pillow`), then runs `python -m pytest ...`" | `.github/workflows/tests.yml` additionally installs `structlog` and `black==23.1.0`, and runs `python -m black --check src tests` as a build-failing gate before the tests. A formatting violation now fails CI, which the instructions do not mention |
[...]

Issue #415, opened 2026-06-07
In-game "download & install update" (Tier 2)


dmccoystephenson:
## Summary
Extend the update notifier (Tier 1) so that, when a newer release exists, the player can update with a single click from inside the game — the game downloads the right release asset and launches the platform installer, instead of sending the user to the browser.

## Behavior
- Reuses the Tier 1 version check (latest GitHub release vs. the build's own version).
- When the player clicks **Update**, the game downloads the matching asset for its platform from the release:
  - **Windows (installed):** `Roam-<version>-Setup.exe` → launch it (optionally `/SILENT`), then exit so it can replace the install.
  - **Windows (portable):** `Roam-<version>-windows-portable.zip` → extract and swap the folder, or hand off to a small helper that does it after the game exits.
  - **macOS:** `Roam-<version>.dmg` → download and open it (the user drags to Applications).
- Show download progress and a clear "the game will close to finish updating" message.

## Why it matters
Removes the manual download/run step from the update flow — the most common friction in "go update the game." User data is unaffected (it lives in `%APPDATA%\Roam` / `~/Library/Application Support/Roam`), so an install-over works cleanly.

## Risks / constraints (important)
- **Executing a downloaded installer is security-sensitive.** The release artifacts are currently **unsigned** (#393 / #396), so auto-download-and-run is exposed to MITM / tampering. At minimum: download over HTTPS from the GitHub release, and **verify a checksum** published with the release before executing. Ideally gate this tier on the installers being code-signed.
- A running process can't replace its own files on Windows while open — the updater must launch the installer and exit, or use a small external helper.
- Keep it **opt-in** and always offer the manual "open Releases page" fallback (Tier 1) in case the auto-flow fails.
- Background download must be cancellable and fail gracefully (offline, partial download, asset missing for this platform).

## Dependencies
Builds on the Tier 1 notifier (in-game update check) and the version-stamping prerequisite. Strongly recommended to land **after** Windows code-signing (#393) and macOS notarization (#396) so the game isn't executing unsigned binaries it fetched.

_Filed by Claude; design grounded in the current Releases pipeline, verified against source._

PR #565, opened 2026-08-14
test: characterization coverage for the agentic (CPC) NPC behavior

### LINKED ISSUE #564: CPC world state renders off-grid cells with the same `?` glyph the legend assigns to creatures
## Summary

The 5×5 ASCII view that `AgenticBehavior._buildWorldState` sends to the model uses `?` for two unrelated things: a tile holding a living entity, and a tile that lies outside the room grid. The legend it ships names only the first, so an NPC standing near a room edge is told it is surrounded by creatures.

## Evidence

Observed against `main` at 353d002.

`src/npc/agenticBehavior.py:229-230` substitutes `?` whenever the coordinate falls off the grid:

```python
loc = room.getGrid().getLocationByCoordinates(x + dx, y + dy)
row.append("?" if loc == -1 else _cellGlyph(loc))
```

`_cellGlyph` (`src/npc/agenticBehavior.py:345-347`) returns the same character for a living entity:

```python
for e in entities:
    if isinstance(e, LivingEntity):
        return "?"
```

and the legend appended to every world state (`src/npc/agenticBehavior.py:238-239`) documents one meaning only:

```
Legend: . empty  T OakWood  S Stone  A Apple  B Banana  W wood-floor  # solid/wall  ? creature
```

For an NPC at grid coordinate `(0, 0)` the entire top row and the entire left column are off-grid, so the serialised view opens with `? ? ? ? ?` and the model is asked to plan against a wall of phantom creatures. The behavior is now pinned by `test_build_world_state_marks_off_grid_cells_with_a_question_mark` in `tests/npc/test_agenticBehavior.py`, which was written as a characterization test rather than as an assertion that the current glyph is correct.

## Impact

Limited to CPC mode (`npcMode: cpc`, which additionally requires `ANTHROPIC_API_KEY`); the deterministic FSM path is unaffected. The consequence is degraded planning near room boundaries, not a crash.

## Suggested resolution

A distinct glyph for an off-grid cell — `X`, for example — added to the legend alongside `? creature`. The existing characterization test would need updating in the same change, since it pins the current output deliberately.

---

---

dmccoystephenson:
## Summary

`src/npc/agenticBehavior.py` was the largest behavior-bearing module in the tree with no test file at all — 360 lines, no `tests/npc/test_agenticBehavior.py` — while its deterministic sibling `ProgrammaticBehavior` has been covered since the feature landed. 57 characterization tests have been added. **No production code is changed by this PR.**

The gap mattered for two reasons. The module is the one place where model output becomes world mutation (`_executeAction` moves the NPC, empties inventory slots and adds entities to the room), and it degrades silently: `_buildClient` returns `None` when `ANTHROPIC_API_KEY` is unset or the `anthropic` package is missing, after which `tick` delegates to `ProgrammaticBehavior`. A regression in the CPC path therefore raises nothing at import time and is invisible to anyone running the default `npcMode: npc`.

## What is covered

- **Client construction** — no API key, package unavailable, and a constructor that raises, all three yielding `None`.
- **FSM fallback** — the `fsm:` state prefix and the goal string are both taken from the fallback, and no AI call is scheduled on that path.
- **Cooldown and queue draining** — a tick inside the movement cooldown is a no-op; exactly one queued action is executed per tick.
- **Call scheduling** — asserted both ways: no second call is scheduled while the `_CALL_INTERVAL_TICKS` interval is unfinished, and one is scheduled once it has elapsed.
- **`_callAI` response handling** — a well-formed response, a markdown-fenced one, actions that are not dicts or lack `type`, an empty goal leaving the previous one intact, invalid JSON, a client that raises, and a response with no text block. `_pendingCall` is asserted clear on every failure path.
- **`_buildWorldState`** — unknown position, the position/energy header, empty and stacked inventories, the 5×5 grid shape, a neighbour's placement relative to the NPC, and off-grid cells.
- **`_cellGlyph`** — every branch, including living-entity precedence and the solid/passable fallthrough.
- **All five action types** — move (including direction wrapping and the room-boundary request), gather, eat, place and idle, each with their no-op paths.

## No network, no sleeping

The Anthropic client is replaced by a `SimpleNamespace` whose `messages.create` returns a canned response object, so the real parsing, queueing and error handling run against controlled input. The daemon thread that `tick` spawns is handled by monkeypatching `agenticBehavior.threading.Thread` with a stand-in that records the target and runs it on `start()` — the captured callback is therefore invoked and `_callAI` genuinely executes, rather than the test asserting only that a thread was created.

## Evidence the tests are not vacuous

An all-green first run on a 57-test file is weak evidence, so three mutations were applied to `src/npc/agenticBehavior.py` and the suite re-run:

| Mutation | Test that failed |
|---|---|
| `self._wantsRoomChange = True` to `False` at the room boundary | `test_execute_action_requests_a_room_change_at_the_boundary` |
| markdown fence strip in `_callAI` disabled | `test_call_ai_strips_markdown_fences` |
| living-entity precedence loop deleted from `_cellGlyph` | `test_cell_glyph_prefers_living_entities` |

Each failed exactly one test and no others. The source was then restored from `git` and the suite reconfirmed green.

## Found, filed, not fixed

#564 — `_buildWorldState` renders an off-grid cell as `?`, the same glyph `_cellGlyph` returns for a living entity, while the legend sent with every world state documents `?` as "creature" only. An NPC at `(0, 0)` opens its view with `? ? ? ? ?` and plans against phantom creatures along two edges. Per the rule that a test-expansion cycle must not change production code, the current output is pinned by `test_build_world_state_marks_off_grid_cells_with_a_question_mark`, commented as characterization rather than approval, and the fix is left to an implementation cycle.

## Deferred this cycle

Every other open issue was left untouched, with the reason recorded here rather than as comments on each: #393/#396 need paid code-signing identities; #415/#416 are explicitly gated on those two; #234 needs an external Viron server that is not running; #550 offers two mutually exclusive resolutions and asks for a decision on which is intended; #551 is a harness-blocked edit to agent-loaded configuration and is on the do-not-auto-merge path list; #556 carries three unanswered design questions about ladder semantics.

## Test plan

- [x] `python3 -m pytest -q --ignore=tests/mapimage/test_mapImageGenerator.py` — 1268 passed, up from 1211 (the ignored file needs Python 3.9+ parenthesized-context-manager syntax that this sandbox's Python 3.8 cannot collect; pre-existing and unrelated, CI runs 3.12 with no exclusion)
- [x] `python3 -m black --check src tests` — clean
- [x] Mutation check above — three source mutations, three distinct failures, source restored
- [x] Nothing under `src/` modified — the diff lists `CHANGELOG.md` and `tests/npc/test_agenticBehavior.py` only

No tracking issue — the gap was found during triage.

---

---


dmccoystephenson:
Self-review rubric — scored by Claude against the diff and command output, not judgment. CI on head `3c3e2ca` is green (`test`, `package`, `package-macos`, `windows-installer`, `Save persistence (Playwright)` all pass).

- **Scope: PASS** — `git diff --name-only origin/main...HEAD` lists exactly `CHANGELOG.md` and `tests/npc/test_agenticBehavior.py`. Nothing under `src/` is touched, and no unrelated formatting or renames appear.
- **Tests-new: PASS** — the diff adds tests only; every public and private surface named in the PR body has at least one test, and the module previously had none.
- **Tests-fix (empirical): PASS** — this is a characterization cycle with no fix to stash, so the equivalent check was run in reverse: three mutations were applied to `src/npc/agenticBehavior.py` (boundary `_wantsRoomChange` inverted, `_callAI` fence strip disabled, `_cellGlyph` living-entity precedence deleted) and each produced exactly one failure — `test_execute_action_requests_a_room_change_at_the_boundary`, `test_call_ai_strips_markdown_fences`, `test_cell_glyph_prefers_living_entities` — with the other 54 still passing. Source restored via `git checkout --` and re-verified green.
- **Sibling structure: PASS** — the new file mirrors `tests/npc/test_programmaticBehavior.py`: the same `_config(tps=30)` / `_setup(...)` helper shape, the same banner-comment section dividers, and the same practice of setting private attributes to arrange state (that file sets `_nearestCache`, `_seekBudget`, `_blockedTargets`).
- **Sibling renames: PASS** — nothing was renamed.
- **Docs: PASS** — no user-visible behavior changed, so `README.md`, `PLANNING.md`, `LOGGING.md`, `config.yml` and `schemas/*.json` are all unaffected; each was checked rather than assumed. `CHANGELOG.md` has a dated `2026-08-13` summary row and an AI Agent Sessions entry.
- **Issue resolution: PASS (vacuously)** — no `Closes #N` is claimed. The gap was found during triage; the one defect discovered while writing the tests was filed as #564 and deliberately not fixed.
- **CI: PASS** — see above.
- **No `@property`: PASS** — `gh pr diff 565 | grep '@property'` returns nothing.
- **DI discipline: PASS** — no new class is introduced into `src/`; the test file resolves nothing from the container and imports no container module, matching the sibling test's direct-construction style for `Room`/`Npc`.
- **No new `print()`: PASS** — nothing under `src/` changed; the test file contains no `print(`.
- **CHANGELOG updated: PASS** — updated, with the Learning Log entry tagged `[not yet integrated]`.
- **Vendored lib untouched: PASS** — no path under `src/lib/` appears in the diff.
- **Markdown fences: PASS** — the `CHANGELOG.md` addition contains no fenced block; the one inline reference to a triple backtick is wrapped in a four-backtick code span.
- **Naming: PASS** — test functions are `test_` + snake_case, helpers and fixtures are camelCase (`_locationOfNpc`, `_stubClient`, `inlineThread`), and the one new class is `PascalCase` (`_InlineThread`).
- **Schema sync: PASS (not applicable)** — no `JsonReaderWriter` serialization changed.

Three observations that did not warrant a change, recorded for the reviewer rather than acted on:

- `tests/npc/test_agenticBehavior.py` — the tests reach into `_client`, `_actionQueue`, `_pendingCall` and `_goalDescription` to arrange and assert state. `AgenticBehavior` exposes no seam for injecting a client, so there is no alternative short of changing production code, which a test-expansion cycle must not do. The sibling `test_programmaticBehavior.py` sets private attributes the same way, so this follows the established convention rather than introducing one.
- `tests/npc/test_agenticBehavior.py` — `test_execute_action_moves_the_npc` asserts only that the location identifier changed, not which neighbour was reached, because `_npcMove` legitimately falls back to a perpendicular direction when the first choice is blocked. Exact destinations are pinned instead by `test_execute_action_wraps_out_of_range_directions` and by the boundary test, so direction handling is not left unverified — but that one assertion is weaker than its neighbours and would not catch an off-by-one in the direction mapping on its own.
- `requirements.txt` does not list `anthropic`, so `_ANTHROPIC_AVAILABLE` is `False` in CI and in this sandbox. The three client-construction tests reach their branches only by monkeypatching the module globals, which means they verify `_buildClient`'s own control flow and not that a real `anthropic.Anthropic(api_key=...)` call signature is still correct. That residual gap is inherent to an optional dependency that is never installed, and is noted rather than papered over.

---
This review comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

